PR #13, opened 2020-01-16
Added Per-Query Configuration override for Xquery

ravi-b-m:
I added the per query configuration for xquery.. 

- Used lodash for finding the block comment

- Did minor modification in `marklogicClient.ts` in function `parseQueryForOverrides`

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ going to 'Security' -> 'Certificate Templates' -> (cert host name), and then
 selecting the "Status" tab. There is an "download" button in the "certificate template status"
 section. Click that button to download a copy of your root CA.
 
-### Per-query configuration override (currently SJS only)
+### Per-query configuration override 
 
 You can override your VS Code configured settings by using a block comment as the first language token
 in the query. The comment should conform to the following:
@@ -107,6 +107,23 @@ e.g.:
 */
 'use strict';
 cts.doc('/my-testing-doc.json');
+```
+
+```xquery
+(:
+/* mlxprs:settings
+{
+  "host": "my-test-host",
+  "contentDb": "unit-test-database",
+  "modulesDb": "unit-test-MODULES",
+  "user": "admin",
+  "pwd": "admin",
+  "note": "Thhese settings are for testing only"
+}
+*/
+:)
+xquery version "1.0-ml";
+fn:doc('/my-testing-doc.json')
 ```
 
 When this query runs, it will use the host, port and contentDb specified in the comment, along with the VS Code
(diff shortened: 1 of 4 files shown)


mikrovvelle:
Looks good at first glance. I will test it locally, and add some automated tests. Thanks!

mikrovvelle (changes requested):
In my tests, lodash isn't recognizing XQuery comments properly. It's picking up the comment opener, `(:` as punctuators.

To reproduce, add the following to `testOverrideQuery.ts`:

```ts
export function testOverrideXQueryWithGoodJSON(): string {
    return `
(: mlxprs:settings

  {
    "host": "overrideHost",
    "port": 12345,
    "username": "overrideUser"
  }

:)
(: Another comment :)
fn:doc(cts:uris()[12 + 19])
`
}
```

And then the following test to `client.test.ts`:

```ts
    test('override parser should recognize XQuery config overrides', () => {
        const queryText: string = testOverrideXQueryWithGoodJSON()
        const overrides = parseQueryForOverrides(queryText)
        assert.equal(overrides.host, 'overrideHost')
        assert.equal(overrides.port, 12345)
    })
```

The test fails for me, and from debugging it appears to be because lodash fails to recognize XQuery comments.

mikrovvelle:
Looking at the README changes now, I see you specify JS-comment-within-an-XQY-comment (`(: /* like this */ :)`. It could work that way, but it might be unwieldy for the user. I'm going to investigate how well that works, and how much work it would be to drop the JS comments requirement for XQuery.

Also, using `_.find(tokens, ['type', 'BlockComment'])` means the overrides could be put elsewhere in the query, not just the top.

ravi-b-m:
> Looking at the README changes now, I see you specify JS-comment-within-an-XQY-comment (`(: /* like this */ :)`. It could work that way, but it might be unwieldy for the user. I'm going to investigate how well that works, and how much work it would be to drop the JS comments requirement for XQuery.
> 
> Also, using `_.find(tokens, ['type', 'BlockComment'])` means the overrides could be put elsewhere in the query, not just the top.

I was not able to find a good tokenizer for xquery like `esprima`.. We can write our own ... due to this I had to rely on `lodash` to find the first block comment.. I  will try to write a custom parser for xquery..

btw: an awesome plugin.. thank you.. this is very useful .. and our ML development is all in xquery.

mikrovvelle:
Thank you! Glad to hear it's useful for you.

You're right that there's not a lot of XQuery support in existing node tools. I've been trying to rig something up with xqlint, but it's not what it's built for. I'm going to see how far I can get with raw string processing. We might be able to avoid importing or writing a whole parser.

PR #29, opened 2020-04-03
JS debugger pre-release

zzzwan:
1. Integrated jsdebugger configurations with mlxpr 
2. Added connectServer/disconnectServer command pairs in extension
3. Sanitized code and fixed some bugs found since last pull request
4. Updated readme

### DIFF
--- a/client/JSDebugger/configurationProvider.ts
+++ b/client/JSDebugger/configurationProvider.ts
@@ -3,15 +3,11 @@
  */
 
 import * as vscode from 'vscode'
-import {DebugConfiguration,WorkspaceFolder, CancellationToken, ProviderResult} from 'vscode'
+import {DebugConfiguration, WorkspaceFolder, CancellationToken, ProviderResult} from 'vscode'
 import * as request from 'request-promise'
+import * as querystring from 'querystring'
 
 export class MLConfigurationProvider implements vscode.DebugConfigurationProvider {
-
-    /**
-	 * Massage a debug configuration just before a debug session is being launched,
-	 * e.g. add all missing attributes to the debug configuration.
-	 */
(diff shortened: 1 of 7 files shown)


client/JSDebugger/configurationProvider.ts
```diff
@@ -119,8 +113,8 @@ export class MLConfigurationProvider implements vscode.DebugConfigurationProvide
         return config
     }
 
-    private async getAvailableRequests(username: string, password: string, servername: string, hostname: string): Promise<any> {
-        const url = `http://${hostname}:8002/jsdbg/v1/paused-requests/${servername}`
+    private async getAvailableRequests(username: string, password: string, debugServerName: string, hostname: string): Promise<string> {
+        const url = `http://${hostname}:8002/jsdbg/v1/paused-requests/${debugServerName}`
```
mikrovvelle, line 117: 1. Will the debug server support https?
2. Should we use a dedicated function instead of building each URL string on the fly?

zzzwan: 1. It should. I need to add the part.
2. Do you mean we have one function that builds url and every other function calls it with parameters?

client/JSDebugger/mlDebug.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
```
mikrovvelle, line 1: because of all the overrides, right?

zzzwan: yes

client/JSDebugger/configurationProvider.ts
mikrovvelle: 1. Ok.
2. Yes. But I'm not sure how much we can generalize that. Let's see how much more complex the code gets while fixing point 1.

PR #73, opened 2020-06-18
sjs adapter testing

### LINKED ISSUE #19: Automated testing
Last-mile integration testing of the extension. This is complementary to the extensive back-end testing that happen as part of the MarkLogic release cycle. 

The minimum we should aim for is a single command that can run all of the tests in one shot. Next step is to hook up CI/CD so that it happens automatically on PRs and releases. (I’ll create a separate issue for that.)

If there are problems with the test library we‘re using we should investigate how others handle this, for example, the extensions that Microsoft delivers.

### LINKED ISSUE #60: Crossing an XQuery eval/invoke boundary
What do we expect the behavior of the debugger to be when trying to debug a request that drops into XQuery? For example, for historical reasons, Data Hub flows and REST Client API Resource Service Extensions are bootstrapped from XQuery, even if the user serviceable code is written in JavaScript. 

I wouldn’t expect the debugger to seamlessly cross over this boundary today. What are our options? 

1. Switch back and forth between interactive XQuery and JavaScript debugging in the same request (I don’t know if that’s even feasible)
2. Continue running the request, but step over any XQuery (Could JavaScript debugging be re-entrant later in the request?)
3. Cleanly abort the request with a sensible message when it tries to cross the XQuery boundary. 

The above is ordered by desirability, descending, but also by complexity/infeasibility. Are there other options? What am I missing?

In the near term and also more generally, I’d suggest extracting the JavaScript from a Data Hub step or a REST Client API Extension into its own library module and debugging it from a unit test (or bootstrap main module). I don’t think we want to encourage debugging framework code in the first place.

zzzwan:


mikrovvelle:
@zzzwan  I've pushed some changes to a dedicate PR branch: 
https://github.com/mikrovvelle/mlxprs/commits/pr73

5 tests won't pass for me. Let's talk on Slack on Monday.

jmakeig:
Remaining issues, probably in implementation order:

* Address reliability of test, e.g. race conditions, inconsistent results
* Separate out unit and integration test

PR #108, opened 2022-11-08
update CONTRIBUTING.md with build instructions

### LINKED ISSUE #107: Update contributing.md to describe how to build and test locally
The current CONTRIBUTING.md file only describes how to submit a pull request. What we need are instructions for the following:

1. How to build and test the plugin locally in VSCode
2. How to run the tests (which I assume should include what's at ./client/test and at ./server/test

The above will allow any developer to review/approve future PR's by manually testing the plugin locally. And we can soon get a Jenkins build together for running the automated tests on every PR and on the `develop` branch when it's updated. 

For bonus points - include some info about what the "client" and "server" folders are for, in particular the latter - I wasn't expecting to see a "server" folder in a VSCode plugin.

bshalke:


CONTRIBUTING.md
```diff
@@ -25,6 +25,52 @@ The following information is most helpful:
...
+    * **extension.ts** — Code insertion point.  Contains initialization code and defines commands callable by user.
+    * **marklogicClient.ts** — Interface to MarkLogic [Node Client API](https://github.com/marklogic/node-client-api) package, which is used to communicate with ML Server.
+    * **marklogic-types.d.ts** — TypeScript type declarations used by the package.
+* **server** — Language server extension folder (provides syntax highlighting; based on vscode-languageserver extension)
+    * **package.json** — As the language server is technically a separate package, this defines the dependencies for it
+    * **server.ts** — Code insertion point.  Contains initialization code for language server.
+
+### Building and Testing Project
+
+Building the project requires [Node.js](https://nodejs.org/) to be installed on your local machine
```
BillFarber: Does it matter which version, or just latest LTS version?

rjrudin (commented):
I'll give this a run-through first thing tomorrow morning

CONTRIBUTING.md
bshalke: That's a great question.  I wrote this based on a few hours of experimentation yesterday so I have no clue.

CONTRIBUTING.md
```diff
@@ -25,6 +25,52 @@ The following information is most helpful:
 * **Pointers** — If you have a hunch or just don't want to bother with a pull request,
 please point me in the right direction (e.g. line numbers) to fix it.
 
+## Development, Builds and Testing
+
+### Project Layout
```
rjrudin, line 30: This is good, but just describe the "client" and "server" files - the other files should be obvious, and we don't want to create an expectation that every top-level file/folder will be documented here.

CONTRIBUTING.md
rjrudin: I'm using Node 14, as @BillFarber  mentioned that we're using that to compile the node-client. Later on, we can test with 16 and 18 as well.

CONTRIBUTING.md
```diff
@@ -25,6 +25,52 @@ The following information is most helpful:
...
+    * **marklogic-types.d.ts** — TypeScript type declarations used by the package.
+* **server** — Language server extension folder (provides syntax highlighting; based on vscode-languageserver extension)
+    * **package.json** — As the language server is technically a separate package, this defines the dependencies for it
+    * **server.ts** — Code insertion point.  Contains initialization code for language server.
+
+### Building and Testing Project
+
+Building the project requires [Node.js](https://nodejs.org/) to be installed on your local machine
+
+First, dev dependencies must be installed.  Run `npm install` in both this folder, and in the `./server` folder.
```
rjrudin: I like using the actual statements to clarify stuff like this when multiple folders are involved - e.g.:

    npm install
    cd server
    npm install
    cd ..

That makes it clear too that `npm install` doesn't need to be run from within the "client" folder (I'm assuming that's the case).

CONTRIBUTING.md
```diff
@@ -25,6 +25,52 @@ The following information is most helpful:
...
+You can choose which environment should be used as the default shell for npm with the command `npm config set script-shell [shell]`.  For instance, to use WSL Bash, you would run `npm config set script-shell bash`.  PowerShell v7+ is `npm config set script-shell pwsh`  You can reset this with `npm config delete script-shell`.  Even if running npm from PowerShell, it will try to use cmd.exe as the default shell if you do not define it.
+
+To compile changes, run `npm run-script compile`.  Note that both the client and the language server need to finish compiling before you attempt to run the extension, or you will get an error.  Other script definitions are in `package.json.`
+
+It is recommended to use VSCode as the editor for this package, as it can self-load a debug instance.
+
+* Open the project folder in VSCode
+* Select the "Run and Debug" sidebar window (Ctrl+Shift+D)
+* Choose "Launch Extension (debug)" from the RUN AND DEBUG dropdown menu
+* Press the green play button or F5 to compile and launch the plugin in a test environment
```
rjrudin, line 72: These steps are really good, but I think we need to tell the developer how to run the equivalent of a "hello world" - otherwise, I'm just looking at a new VS Code instance with a blank screen, wondering what to do. 

For example, I opened a new editor and entered `console.log("Hi")` and then knew to select "MarkLogic: Eval JS" from the command palette, but I get this error:

```
{
  "message": "eval XQuery on server: cannot process response with 500 status",
  "statusCode": 500,
  "body": {
    "errorResponse": {
      "statusCode": 500,
      "status": "Internal Server Error",
      "messageCode": "INTERNAL ERROR",
      "message": "XDMP-RWEMPTY: The value of expression '$0' is required to be non-empty in rule: set-database . See the MarkLogic server error log for further detail."
    }
  }
}
```

I checked that my default ML settings - localhost/8000/admin/admin - are valid. I tried highlighting the text and calling the command, but no luck. 

Hopefully this is just a config issue for me, but regardless - this guide should provide a "hello world" example so that the developer knows their test environment is working correctly.

rjrudin (changes requested):
Thanks for getting this going @bshalke . Main request is to provide "hello world" instructions once the test environment is launched. But this is a big step in the right direction.

CONTRIBUTING.md
bshalke: It is.  The install should be run from the same folder as the package files, which are in root and server respectively.  I'll make that more explicit

CONTRIBUTING.md
bshalke: I think it's just a config issue because it worked right away for me.  The documentation on the config lies and it does actually require you to set the database name instead of relying on the default of the app server, as far as I can tell. I think that should be fixed in the readme and we should point the user there for initial config, since the readme should probably contain a working hello-world example.

bshalke: I just checked, and yes - it's a bug.  The Documents database needs to be set.  I'll add a note in the Readme, remove the misleading text in the config options and create a new issue for it.

bshalke: https://github.com/marklogic-community/mlxprs/issues/109

CONTRIBUTING.md
```diff
@@ -45,9 +40,14 @@ please point me in the right direction (e.g. line numbers) to fix it.
...
-Building the project requires [Node.js](https://nodejs.org/) to be installed on your local machine
+Building the project requires [Node.js](https://nodejs.org/) to be installed on your local machine.  Node v14 LTS is recommended.
 
-First, dev dependencies must be installed.  Run `npm install` in both this folder, and in the `./server` folder.
+First, dev dependencies must be installed.  Run `npm install` in both this folder, and in the `./server` folder.  For example:
+
+`npm install`
+`cd server`
+`npm install`
+`cd ..`
```
rjrudin: Markdown is picky about this - the above results in all 4 commands on one line. You can either indent each line and not use the tick marks, or just do this, which is to put the commands on separate lines and then 3 tick marks on the line before and after those lines:

```
npm install
cd server
npm install
cd ..
```

rjrudin (changes requested):
This LGTM, just need that one tweak on the 4 commands. Please squash the commits into one commit and then I'll approve and merge.

CONTRIBUTING.md
bshalke: Done

CONTRIBUTING.md
bshalke: Since it just caught me up - the version of @types/node in master is 14.14.5.  I just updated that to 14.17.0 in a different branch I'm working on so I could use crypto.randomUUID().  Latest version of 14 is 14.21.1, and that's what I'm using.

CONTRIBUTING.md
```diff
@@ -25,6 +25,55 @@ The following information is most helpful:
...
+
+```
+npm install
+cd server
+npm install
+cd ..
+```
+
+***IMPORTANT NOTE:***
+*The shell scripts used to drive node compilation assume a \*nix environment.  If you are on Windows, you have a few options:*
```
rjrudin, line 55: I can't verify these instructions, but I'll assume they're true. :) We can always update these if/when someone runs into problems while trying to contribute from Windows.

rjrudin (approved):
This LGTM; we can update the section about running tests once we have them all passing, along with a Jenkins job for them.

Issue #112, opened 2022-11-14
Client tests are failing


rjrudin:
I'm getting 2 failures out of 17; the error messages seem to be:

```
1) Extension Test Suite
       getDbClient should cache its settings in the state passed to it:
at Context.<anonymous> (/Users/rrudin/workspace/mlxprs/client/test/suite/client.test.ts:38:16)
```

And:

```
2) Extension Test Suite
       overrides should not touch parameters they do not specify:
     AssertionError [ERR_ASSERTION]: 'undefined' == undefined
      at Context.<anonymous> (/Users/rrudin/workspace/mlxprs/client/test/suite/client.test.ts:59:16)
```

Need to get these passing so that we can then get a Jenkins job setup and require the tests to pass on future PR's.

PR #113, opened 2022-11-15
#112 Fixing 2 failing tests

rjrudin:
Did a little cleanup while I was at it:

- Removed fields in `launch.json` that VSCode said are unused / have no effect
- Fixed warnings in client.test.ts so that e.g. `strictEqual` is used instead of `equal`
- Replaced erroneous occurrences of "MarkLogicVSClient"

### DIFF
--- a/client/marklogicClient.ts
+++ b/client/marklogicClient.ts
@@ -36,6 +36,12 @@ export class MlClientParameters {
         this.ssl = Boolean(rawParams.ssl)
         this.pathToCa = rawParams.pathToCa || ''
         this.rejectUnauthorized = Boolean(rawParams.rejectUnauthorized)
+
+        // This check was previously done in the MarklogicClient constructor, but doing so causes the sameAs
+        // function in this class to not behave properly
+        if (this.authType !== 'DIGEST' && this.authType !== 'BASIC') {
+            this.authType = 'DIGEST'
+        }
     }
 
     toString(): string {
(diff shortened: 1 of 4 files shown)


client/marklogicClient.ts
```diff
@@ -56,7 +62,7 @@ export class MlClientParameters {
     sameAs(other: MlClientParameters): boolean {
         return (
             this.host === other.host &&
-            this.port === other.port &&
+            ((isNaN(this.port) && isNaN(other.port)) || this.port === other.port) &&
```
bshalke, line 65: I usually prefer to do ports as strings instead of numbers for reasons like this.  Is there a reason to use numbers here?

rjrudin: It's a property of the `MlClientParameters` class - I didn't want to try changing the class definition to make the test impact, given the larger blast radius of doing so. Just wanted to get the test passing, and that test for some reason has port=NaN, so this check is needed since NaN !== NaN.

bshalke (commented):
I was able to get around to testing these and both the client and server tests pass, however I have 4 failing JS Debug tests.  Not sure if you were checking for those or not.

[jsdebugresults.txt](https://github.com/marklogic-community/mlxprs/files/10037622/jsdebugresults.txt)

rjrudin:
@bshalke My general preference is for slim / narrowly-scoped issues and PRs and incrementally moving things forward, so I think lets keep this PR to just the "client" tests and use a separate one for the "js debug" tests (which I was not aware of).

bshalke:
> @bshalke My general preference is for slim / narrowly-scoped issues and PRs and incrementally moving things forward, so I think lets keep this PR to just the "client" tests and use a separate one for the "js debug" tests (which I was not aware of).

That works.  I'll approve and open a new issue for those tests

bshalke (approved):

PR #120, opened 2022-12-27
SJS file associations

### LINKED ISSUE #119: File Associations for MJS and SJS
By default MJS and SJS should have the correct formatter associated with the file types.

bshalke:
Adds file association for .sjs files to javascript language templates

Resolves #119

### DIFF
--- a/package.json
+++ b/package.json
@@ -388,6 +388,12 @@
                     ".xquery"
                 ],
                 "configuration": "./xquery-ml.configuration.json"
+            },
+            {
+                "id": "javascript",
+                "extensions": [
+                    ".sjs"
+                ]
             }
         ],
         "grammars": [


package.json
```diff
@@ -388,6 +388,12 @@
                     ".xquery"
                 ],
                 "configuration": "./xquery-ml.configuration.json"
+            },
+            {
+                "id": "javascript",
```
rjrudin, line 393: Is this telling VSCode - hey, when you use this plugin, use the existing "javascript" language for files ending in ".sjs"?

bshalke: It's just associating the .sjs file extension with that language ID (which is the ID for the built-in JavaScript language). So yes, except that there's no requirement that the language has to already exist. It just happens to in this case.

rjrudin (approved):

PR #126, opened 2023-03-03
DEVEXP-121: documentsDb does not correctly default to app server settings when left empty

BillFarber:


client/test/suite/defaultDb.test.ts
```diff
@@ -0,0 +1,55 @@
...
+suite('Default Documents DB Test Suite', () => {
+    test('When MlClientParameters is used and the default contentDb is null', async () => {
+        const mlClient = new MarklogicClient(
+            new MlClientParameters({
+                host: 'host',
+                port: 'port',
+                user: 'user',
+                pwd: 'pwd',
+                authType: 'DIGEST',
+                contentDb: null,
```
rjrudin, line 13: Nice tests! I am thinking, strictly out of paranoia, that it's worth testing the following as well:

    contentDb: undefined
    contentDb: ""

According to https://stackoverflow.com/questions/154059/how-do-i-check-for-an-empty-undefined-null-string-in-javascript , you may be able to simplify the code to just do:

    if (!this.params.contentDb) {
      this.params.contentDb = null;
    }

So basically normalize it to null if it's a "falsey" value.

rjrudin (approved):

PR #127, opened 2023-03-03
DEVEXP-120: Document how to run client tests via command line

BillFarber:


BillFarber:
The instructions don't include this, but it might be worth starting from a clean slate, so I've been doing something like the following before testing:
rm -rf  node_modules/ .vscode-test/ dist/ server/dist/ server/node_modules/ server/out/

CONTRIBUTING.md
```diff
@@ -119,10 +119,17 @@ The test script will perform following:
...
-* npm test - runs the client tests (not including the integration tests)
-* npm run testServer - runs the server tests
-* There's nothing in place yet for running the integration tests from the command line
-
+Follow these steps to run the tests from the command line.
+* npm install
+* cd server/
+* npm install
+* cd ..
+* npm run-script compile
```
rjrudin: Running these 5 commands seems common enough that I think it'd be nice to add a new "script" to package.json that runs all 5 of them - e.g. 

    npm run-script compile && npm run pretest && ...

Could call that "runClientAndServerTests" or maybe something shorter that conveys the same thing. 

I would leave the integration tests out for now since those are more fragile.

compileAndPrepForTests.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+# Uncomment the following line to delete all artifacts and build from scratch
+# rm -rf node_modules/ .vscode-test/ dist/ server/dist/ server/node_modules/ server/out/
+
+npm install
```
rjrudin: Can this just be a script in package.json instead? My impression is that's the convention for Node projects, much like making custom tasks in build.gradle. Same goes for the other .sh file as well. That also keeps things platform-independent.

package.json
```diff
@@ -453,7 +457,8 @@
         "yargs-parser": "^20.2.3"
     },
     "dependencies": {
+        "rimraf": "4.3.0",
```
rjrudin, line 460: My favorite!

rjrudin (approved):

package.json
BillFarber: I did it just for you. Happy birthday.

PR #128, opened 2023-03-06
Refactoring the client integration tests

BillFarber:
This refactoring probably isn't perfect yet, but I think it's a good start.
The 3 tests still fail under certain conditions. They may need to be rewritten, but I think this refactor will help make rewriting easier.


rjrudin (commented):
I don't know enough yet to comment on these changes - I'm good with them if @bshalke is good with them.

PR #131, opened 2023-03-09
Handle no paused requests better

BillFarber:
When the user attempts to "Attach to Remote JavaScript Request" the user is presented with a list of remote requests to choose from.

However, if there are no paused requests, the list is blank, but it is not apparent that VSC is still waiting for the user to "choose" something.

This change causes a message to appear telling the user that no requests were found, and VSC does not continue "waiting".

I also fix a mistake I made in the contributing file.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -9,5 +9,7 @@
     "editor.formatOnSave": true,
     "[typescript]": {
         "editor.defaultFormatter": "vscode.typescript-language-features"
-    }
+    },
+    "marklogic.port": 8055,
+    "marklogic.rejectUnauthorized": false
 }
--- a/client/JSDebugger/configurationProvider.ts
+++ b/client/JSDebugger/configurationProvider.ts
@@ -53,7 +53,7 @@ export class MLConfigurationProvider implements vscode.DebugConfigurationProvide
     }
 
     /* helper function to resolve config parameters */
-    private async resolveRemainingDebugConfiguration(folder: WorkspaceFolder | undefined, config: DebugConfiguration, _token?: CancellationToken): Promise<DebugConfiguration>  {
+    private async resolveRemainingDebugConfiguration(folder: WorkspaceFolder | undefined, config: DebugConfiguration, _token?: CancellationToken): Promise<DebugConfiguration> {
         // acquire extension settings
         const wcfg: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration()
         config.hostname = String(wcfg.get('marklogic.host'))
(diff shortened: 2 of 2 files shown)


client/JSDebugger/configurationProvider.ts
```diff
@@ -139,13 +141,18 @@ export class MLConfigurationProvider implements vscode.DebugConfigurationProvide
...
+                if (!item) {
+                    return vscode.window.showErrorMessage('Request not selected').then(() => {
+                        return undefined	// abort
+                    })
+                }
+                config.rid = item.label
+            } else {
+                vscode.window.showErrorMessage('No paused requests found on server')
+                return undefined
             }
```
BillFarber, line 155: The above is the main purpose of this PR.
Everything in this file is formatting/ESLint-related.

.vscode/settings.json
```diff
@@ -9,5 +9,7 @@
     "editor.formatOnSave": true,
     "[typescript]": {
         "editor.defaultFormatter": "vscode.typescript-language-features"
-    }
+    },
```
rjrudin, line 12: Does this file only affect folks developing mlxprs? Or is it part of the published plugin?

CONTRIBUTING.md
```diff
@@ -100,7 +100,7 @@ Ideally, you should use a dedicated MarkLogic instance for this purpose altogeht
 
 On the MarkLogic instance, create an HTTP server in MarkLogic with the following non-default properties:
 
-- server name: JSdebugTestServer
+- server name: localhost
```
rjrudin: If this should be a host name instead of a server name, can you rename this to "host name" (or just "host")?

client/JSDebugger/configurationProvider.ts
```diff
@@ -139,13 +141,18 @@ export class MLConfigurationProvider implements vscode.DebugConfigurationProvide
                     })
                 }
             }
-            const item = await vscode.window.showQuickPick(items, { placeHolder: 'Select the request to attach to' })
-            if (!item) {
-                return vscode.window.showErrorMessage('Request not selected').then(() => {
-                    return undefined	// abort
-                })
+            if (items.length > 0) {
```
rjrudin: I recommend renaming `items` to `pausedRequests`, which will also allow for getting rid of that "query for paused requests" comment.

.vscode/settings.json
BillFarber: Yes, local development only.

CONTRIBUTING.md
BillFarber: ack. Clearly I was sleepy. The original is correct. I will revert this change.

rjrudin (approved):

PR #132, opened 2023-03-10
DEVEXP-235: Bug when eval'ing a tab

BillFarber:
The circular object reference error that is report is coincidental to the actual problem.
Instead, I fixed the real cause preventing the first eval request from succeeding.

### DIFF
--- a/client/marklogicClient.ts
+++ b/client/marklogicClient.ts
@@ -9,17 +9,17 @@ const SJS = 'sjs'
 export const XQY = 'xqy'
 
 export class MlClientParameters {
-    contentDb: string;
-    modulesDb: string;
+    contentDb: string
+    modulesDb: string
 
-    host: string;
-    port: number;
-    user: string;
-    pwd: string;
-    authType: string;
-    ssl: boolean;
-    pathToCa: string;
-    rejectUnauthorized: boolean;
+    host: string
+    port: number
+    user: string
+    pwd: string
+    authType: string
+    ssl: boolean
+    pathToCa: string
+    rejectUnauthorized: boolean
     /**
      * note: defaults not applied here. Properties can remain undefined so that
      *       per-query overrides don't clobber the existing config with default values.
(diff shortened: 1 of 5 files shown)


client/marklogicClient.ts
```diff
@@ -76,10 +73,10 @@ export class MlClientParameters {
 }
 
 export class MarklogicClient {
-    params: MlClientParameters;
-    ca: string;
+    params: MlClientParameters
+    ca: string
 
-    mldbClient: ml.DatabaseClient;
+    mldbClient: ml.DatabaseClient
```
rjrudin, line 79: This comment is out of scope of this PR - but in general, I think we should use the term "client" or "databaseClient" when referring to an instance of either the Node or Java client. I know there was a movement early on to use the term "db" to emphasize that MarkLogic is a database - but I don't think that would be intuitive to a developer. The developer knows they're making HTTP calls to MarkLogic, and they're accustomed to using a "client" to make HTTP calls. Or even if they're thinking in terms of "I'm using the Java/Node SDK", they're accustomed to "I will use the client object in this SDK to talk to the server". 

Additionally, the name "DatabaseClient" itself suggests that a good var name is simply "client" (for example, for a class name "DatabaseReplicator", it would not make sense to call that "db" instead of "replicator").

So no action, I'm just pontificating - but I think it's good to rename this stuff when we see it (though not in this PR), assuming it's not part of a public API.

client/queryResultsContentProvider.ts
```diff
@@ -79,7 +79,11 @@ export class QueryResultsContentProvider implements TextDocumentContentProvider
         } else {
             window.showInformationMessage(`Query in ${uri.query} got an empty response from ${uri.authority}`)
         }
-        const responseUri = Uri.parse(`${QueryResultsContentProvider.scheme}://${uri.authority}${uri.path}.${fmt}?${uri.query}`)
+        let validUriPath = uri.path
+        if (validUriPath.startsWith('//')) {
```
rjrudin, line 83: Nice - do you know how the URI path was ending up with "//" in the path?

rjrudin (approved):

client/marklogicClient.ts
BillFarber: There are a TON of style-related things that need to be refactored. Some highlighted by ESLint, some not. I need to create a story to one big style cleanup.

client/queryResultsContentProvider.ts
BillFarber: There's another function that builds the URI initially. It is adding a "/" because sometimes the next value doesn't have one. I probably could have changed the URI building there. I'm not sure there's a difference either way.

PR #133, opened 2023-03-13
Create build.gradle to generate test ML App Server

BillFarber:


gradle.properties
```diff
@@ -0,0 +1,8 @@
+mlAppName=JSdebugTestServer
+mlHost=localhost
+mlUsername=admin
+mlPassword=changeme-in-gradle-local.properties
+mlRestPort=8055
+mlContentDatabaseName=Documents
```
rjrudin: I am hoping it's an anti-pattern that Documents/Modules are used, and that was just done before so a developer didn't have to create these databases as well. Can you try removing this and the modules property below so that you get new content/modules databases? That should simplify things and avoid any issues with what someone already has in those databases.

rjrudin: Hmm - I just did a search on `'Modules` and saw that some tests refer to that. Looks like it should be simple to change that to `JSdebugTestServer-modules` - though as long as we're changing this, let's use "mlxprs-test" as the name instead so someone can look at the Admin UI and have an idea of what the purpose of the app server is.

rjrudin (changes requested):
The CONTRIBUTING should be updated too - the "Setup" section can just say to run "./gradlew -i mlDeploy" (and that Java 8 is required).

rjrudin:
Should also include gradlew + gradlew.bat + the ./gradle/wrapper directory so that users don't need to install Gradle.

CONTRIBUTING.md
```diff
@@ -102,61 +115,44 @@ For the Client-Integration test, if you need to override the default properties,
...
-- npm run npmInstallClientAndServer
-- export ML_PASSWORD=\<password> ;npm run testAll
-
+```
+npm run npmInstallClientAndServer
+npm run test
+npm run testServer
+export ML_PASSWORD=\<password> ;npm run testIntegration
+```
+All three test sets may also be run with a single command:
```
rjrudin, line 127: 🌶️

rjrudin (approved):
LGTM, big step in a friendlier direction

PR #137, opened 2023-03-14
DEVEXP-256: handle error better

BillFarber:
No testing for this as the return value is handled by VSCode internals.

Handle the error properly so launching can be re-tried.
Clarify the error message
Clarify the settings description.

### DIFF
--- a/package.json
+++ b/package.json
@@ -302,7 +302,7 @@
                 "marklogic.documentsDb": {
                     "type": "string",
                     "default": "",
-                    "description": "database against which to evaluate queries; required"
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -302,7 +302,7 @@
                 "marklogic.documentsDb": {
                     "type": "string",
                     "default": "",
-                    "description": "database against which to evaluate queries; required"
+                    "description": "database against which to evaluate queries; required for debugging a local script"
```
rjrudin, line 305: HOT

rjrudin (approved):

PR #139, opened 2023-03-15
Integration tests are more reliable now.

BillFarber:
I'm starting to think that .vscode/settings.json should not be checked in since it needs to have the developer's ML password.

### DIFF
--- a/client/test/integration/index.ts
+++ b/client/test/integration/index.ts
@@ -2,22 +2,28 @@ import * as path from 'path'
 import * as Mocha from 'mocha'
 import * as glob from 'glob'
 import { IntegrationTestHelper } from './markLogicIntegrationTestHelper'
+import * as vscode from 'vscode'
 
-export function run(): Promise<void> {
+export async function run(): Promise<void> {
     const mocha = new Mocha({
         ui: 'tdd'
     })
     globalThis.integrationTestHelper = new IntegrationTestHelper()
+    await globalThis.integrationTestHelper.beforeEverything()
+
+    // Without extending the setup, only 1 reporter is support.
+    // For available reporters, see https://mochajs.org/#reporters
+    // For options on configuring multiple reports, see:
+    // https://github.com/mochajs/mocha/pull/1360
+    // https://github.com/glenjamin/mocha-multi
+    const resultsFileName = __dirname + '/../../../results/integrationTestResults.xml'
+    mocha.reporter('xunit', { output: resultsFileName })
+    // mocha.reporter('spec')
+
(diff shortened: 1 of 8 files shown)


client/test/integration/index.ts
```diff
@@ -2,38 +2,48 @@ import * as path from 'path'
...
             globalThis.integrationTestHelper.setupEachTest()
         },
-        afterAll: async () => {
-            await globalThis.integrationTestHelper.afterEverything()
-        },
         afterEach: () => {
             globalThis.integrationTestHelper.teardownEachTest()
         }
     })
+    // afterAll: async () => {
```
rjrudin: Can this be deleted instead of commented out?

package.json
```diff
@@ -418,7 +418,7 @@
     "scripts": {
         "completeClean": "rimraf node_modules/ .vscode-test/ dist/ server/dist/ server/node_modules/ server/out/ mlxprs-${version}.vsix",
         "npmInstallClientAndServer": "npm install && cd server && npm install && cd ..",
-        "testAll": "npm run test && npm run testIntegration && npm run testServer",
+        "testAll": "npm run test && npm run testServer",
```
rjrudin, line 421: Does `testIntegration` no longer need to be run?

client/test/integration/index.ts
BillFarber: It can be. I'm just not happy with it yet. I'd really like to delete the test files, but the delete function isn't reliable. So, my plan is to come back to this sometime soon.

package.json
BillFarber: It does not. In fact, testIntegration does not work work properly.

bshalke:
> I'm starting to think that .vscode/settings.json should not be checked in since it needs to have the developer's ML password.

I agree with this, unless there's a way to have both a settings.json and a my-settings.json that overrides it or something.

BillFarber:
> I agree with this, unless there's a way to have both a settings.json and a my-settings.json that overrides it or something.

How about if I go ahead and delete the file, but put an example of the file in the CONTRIB document?

bshalke:
> > I agree with this, unless there's a way to have both a settings.json and a my-settings.json that overrides it or something.
> 
> How about if I go ahead and delete the file, but put an example of the file in the CONTRIB document?

I think the problem with that is people will create the file and accidentally commit it with their password.  Maybe instead, we should leave it but add it to the gitignore so by default, it won't commit any changes the user makes?  I feel like there's probably an established best practice for this already.

rjrudin (approved):

Issue #116, opened 2022-12-02
Update as many dependencies as possible


bshalke:
There are currently many dependencies that are out of date and have vulnerabilities.  We should update as many of these as we can before release.  Several of these are critical vulnerabilities:

  Critical        Prototype Pollution in minimist                               
  Package         minimist                                                      
  Dependency of   vscode-debugadapter [dev]                                     
  Path            vscode-debugadapter > mkdirp > minimist                       
  More info       https://github.com/advisories/GHSA-xvch-5gv4-984h     

  Critical        json-schema is vulnerable to Prototype Pollution              
  Package         json-schema                                                   
  Dependency of   request [dev]                                                 
  Path            request > http-signature > jsprim > json-schema               
  More info       https://github.com/advisories/GHSA-896r-f27r-55mw 
  
  Critical        Prototype pollution in webpack loader-utils                   
  Package         loader-utils                                                  
  Dependency of   ts-loader [dev]                                               
  Path            ts-loader > loader-utils                                      
  More info       https://github.com/advisories/GHSA-76p3-8jx3-jpfq

mitchshepherd:
Addressed in MLXPRS 3.6.0 release

PR #141, opened 2023-03-16
Updates to the changelog and readme in prep for 2.6.0

BillFarber:


CHANGELOG.md
```diff
@@ -1,3 +1,18 @@
+## [3.6.0](https://github.com/marklogic-community/mlxprs/releases/tag/v3.6.0)
```
rjrudin, line 1: Let's adopt the approach of the Java Client here - https://github.com/marklogic/java-client-api/blob/master/CHANGELOG.md - and only store these in the actual Release page. 

I've found this to be really useful for the ml-gradle projects, because it's convenient to be able to edit these without having to make a commit. 

I just created a new 3.6.0 draft release - just paste the good content you have here into that release description.

CHANGELOG.md
```diff
@@ -1,3 +1,18 @@
...
+
+- [#116](https://github.com/marklogic-community/mlxprs/issues/116) Updated dependencies and fixed all known security vulnerabilities
```
rjrudin, line 3: You don't need a link - just do "#116" and GitHub will automatically create a link for you.

CHANGELOG.md
```diff
@@ -1,3 +1,18 @@
...
+- [#119](https://github.com/marklogic-community/mlxprs/issues/119) Added SJS binding to javascript language
```
rjrudin: javascript -> JavaScript

CHANGELOG.md
```diff
@@ -1,3 +1,18 @@
...
+- [#97](https://github.com/marklogic-community/mlxprs/issues/97) Added a configuration option for the MarkLogic Manage AppServer port (used for the SJS debug server port)
```
rjrudin: Sorry, I'm being picky - "Manage app server".

CHANGELOG.md
```diff
@@ -1,3 +1,18 @@
...
+- [#117](https://github.com/marklogic-community/mlxprs/issues/117) Fixed a bug with the first attempt to eval a JavaScript tab
+- [#109](https://github.com/marklogic-community/mlxprs/issues/109) Prevent the database param from being used in REST calls when the database setting is empty
+- Many updates to the project to facility future enhancements and maintenance
```
rjrudin: Release notes should only be for users, so remove this section since it's intended for developers. Makes it easy for users to scan release notes to find items of interest.

rjrudin (changes requested):

CHANGELOG.md
BillFarber: Leaving CHANGELOG.md in place since VSCode, both the marketplace and the application, are expecting it.

PR #142, opened 2023-03-19
DEVEXP-265: First attempt to eval a SPARQL tab results in an error

BillFarber:


client/queryResultsContentProvider.ts
```diff
@@ -97,10 +97,14 @@ export class QueryResultsContentProvider implements TextDocumentContentProvider
...
      * @returns Promise responseUri where VS Code will retrieve the content to show @async
      */
-    public async writeSparqlResponseToUri(uri: Uri, response: Record<string, unknown>): Promise<Uri> {
+    public async writeSparqlResponseToUri(uri: Uri, response: Record<string, any>): Promise<Uri> {
         const contentType: contentType = workspace.getConfiguration().get('marklogic.sparqlContentType')
-        const fmt: string = contentType.toLowerCase().replace(/^\w+\//, '')
-        const responseUri = Uri.parse(`${QueryResultsContentProvider.scheme}://${uri.authority}${uri.path}.${fmt}?${uri.query}`)
+        const fmt: string = (contentType) ? contentType.toLowerCase().replace(/^\w+\//, '') : 'json'
+        let validUriPath = uri.path
+        if (validUriPath.startsWith('//')) {
```
rjrudin: There are 3 occurrences in this class of this check along with the following Uri.parse line - I think the 3 occurrences should be condensed into something like this (it should be private if that's possible):

```
buildResponseUri(uri: Uri, uriSuffix: String) {
  let validUriPath = uri.path
  if (validUriPath.startsWith('//')) {
    validUriPath = validUriPath.substring(1)
  }
  return Uri.parse(`${QueryResultsContentProvider.scheme}://${uri.authority}${validUriPath}.${uriSuffix}?${uri.query}`)
}
```

rjrudin (commented):
LGTM, added an idea for refactoring.

client/queryResultsContentProvider.ts
BillFarber: Funny, I actually realized this last night during a moment of sleeplessness.
Fixed.

PR #143, opened 2023-03-22
Submit an Optic DSL Query to /v1/rows

BillFarber:


client/marklogicClient.ts
```diff
@@ -189,12 +189,40 @@ export function sendXQuery(
...
+
+export function sendRows(db: MarklogicClient, actualQuery: string): Promise<ml.RowsResponse> {
+    if (actualQuery.startsWith('{')) {
+        return sendRowsSerialized(db, actualQuery)
+    } else {
+        return db.mldbClient.rows.query(actualQuery, { 'queryType': 'dsl' })
+    }
+}
+
+function sendRowsSerialized(db: MarklogicClient, actualQuery: string): Promise<ml.RowsResponse> {
```
rjrudin, line 208: Hmmm.... lot of code for something that I don't think will ever get used... or did Github CoPilot write this? ;)

rjrudin (approved):
Tested out locally, worked well, including errors

client/marklogicClient.ts
BillFarber: During my time with GraphQL, I have needed to experiment with serialized queries a few times. This could have come in handy.

PR #145, opened 2023-03-23
DEVEXP-262: Submit Optic Query and get back XML or CSV

BillFarber:
TypeScripting it up, baby

### DIFF
--- a/client/extension.ts
+++ b/client/extension.ts
@@ -58,14 +58,22 @@ export function activate(context: vscode.ExtensionContext): void {
         const uri = QueryResultsContentProvider.encodeLocation(editor.document.uri, host, port);
         editorSparqlQuery(client, actualQuery, uri, editor, provider);
     });
-    const sendRowsQuery = vscode.commands.registerTextEditorCommand('extension.sendRowsQuery', editor => {
+
+    function sendEditorRowsQuery(editor, rowsResponseFormat: ml.RowsResponseFormat) {
         const actualQuery: string = editor.document.getText();
         const cfg: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration();
         const client: MarklogicClient = cascadeOverrideClient('', SJS, cfg, context.globalState);
         const host = client.params.host; const port = client.params.port;
         const uri = QueryResultsContentProvider.encodeLocation(editor.document.uri, host, port);
-        editorRowsQuery(client, actualQuery, uri, editor, provider);
-    });
+        editorRowsQuery(client, actualQuery, uri, editor, provider, rowsResponseFormat);
+    }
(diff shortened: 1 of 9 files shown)


client/extension.ts
```diff
@@ -58,13 +58,29 @@ export function activate(context: vscode.ExtensionContext): void {
...
+    const sendRowsJsonQuery = vscode.commands.registerTextEditorCommand('extension.sendRowsJsonQuery', editor => {
         const actualQuery: string = editor.document.getText();
         const cfg: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration();
         const client: MarklogicClient = cascadeOverrideClient('', SJS, cfg, context.globalState);
         const host = client.params.host; const port = client.params.port;
         const uri = QueryResultsContentProvider.encodeLocation(editor.document.uri, host, port);
-        editorRowsQuery(client, actualQuery, uri, editor, provider);
+        editorRowsQuery(client, actualQuery, uri, editor, provider, 'json');
+    });
+    const sendRowsCsvQuery = vscode.commands.registerTextEditorCommand('extension.sendRowsCsvQuery', editor => {
```
rjrudin: To avoid the duplication, I think you can make a function that accepts the format arg and then just do something like:

    registerTextEditorCommand('...', editor => sendEditorRowsQuery(editor, "json"))

Where `sendEditorRowsQuery` would have these repeated 6 lines.

client/marklogic-types.d.ts
```diff
@@ -8,6 +8,8 @@ declare module 'marklogic' {
     }
 
     export type contentType = 'application/json' | 'application/xml' | 'text/html' | 'text/csv'
+    export type ResponseFormat = 'json' | 'xml' | 'csv'
```
rjrudin: Since you have `RowsOptions` below - which is good, since it's specific to the rows endpoint - I think these should be `RowsResponseFormat` and `RowsQueryType` since both are specific to the rows endpoint as well.

rjrudin (changes requested):

client/extension.ts
```diff
@@ -58,14 +58,22 @@ export function activate(context: vscode.ExtensionContext): void {
         const uri = QueryResultsContentProvider.encodeLocation(editor.document.uri, host, port);
         editorSparqlQuery(client, actualQuery, uri, editor, provider);
     });
-    const sendRowsQuery = vscode.commands.registerTextEditorCommand('extension.sendRowsQuery', editor => {
+
+    function sendEditorRowsQuery(editor, rowsResponseFormat: ml.RowsResponseFormat) {
```
rjrudin, line 62: 🔥

rjrudin (approved):

PR #147, opened 2023-03-24
DEVEXP-353: View "Connected" status information

BillFarber:
This is hot.

### DIFF
--- a/client/marklogicClient.ts
+++ b/client/marklogicClient.ts
@@ -2,6 +2,7 @@
 
 import * as ml from 'marklogic';
 import * as fs from 'fs';
+import { MlxprsStatus } from './mlxprsStatus';
 
 export const MLDBCLIENT = 'mldbClient';
 export const MLSETTINGSFLAG = /mlxprs:settings/;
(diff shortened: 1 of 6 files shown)


client/marklogicClient.ts
```diff
@@ -106,6 +107,32 @@ export class MarklogicClient {
     hasSameParamsAs(newParams: MlClientParameters): boolean {
         return this.params.sameAs(newParams);
     }
+
+
+    evalSimpleXQueryScript(xQueryScript: string): ml.ResultProvider<ml.Item[]> {
```
rjrudin: I don't think you need this - I think you can just call `mldbClient.xqueryEval` and all will work (the declaration of the ML XQuery version should rarely matter, and if it does for a particular script, I think that script should declare it - I've call eval from the Java Client many times, I don't recall ever once declaring the XQuery version). 

This avoids having to define what "simple" means as well - i.e. I could pass a very complex XQuery script into this function.

client/marklogicClient.ts
```diff
@@ -106,6 +107,32 @@ export class MarklogicClient {
...
+
+    async getConnectedServers(requestingObject: MlxprsStatus, updateCallback: (connectedServers: string[]) => void): Promise<void> {
+        return this.evalSimpleXQueryScript('dbg:connected()')
+            .result(
+                (items: ml.Item[]) => {
+                    const connectedServers: string[] = [];
+                    items.forEach((item) => {
+                        connectedServers.push(item.value);
+                    });
+                    // requestingObject['updateStatusBarItem'](connectedServers);
```
rjrudin, line 120: Can these commented lines be removed?

client/mlxprsStatus.ts
```diff
@@ -0,0 +1,55 @@
...
+    private statusBarItem: vscode.StatusBarItem;
+    private commandId = 'sample.showSelectionCount';
+    private command: vscode.Disposable;
+    private connectedServers = null;
+
+    constructor(context: vscode.ExtensionContext) {
+        const cfg: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration();
+        this.mlClient = cascadeOverrideClient('', SJS, cfg, context.globalState);
+        this.statusBarItem = vscode.window.createStatusBarItem(vscode.StatusBarAlignment.Left, 100);
+        this.command = vscode.commands.registerCommand(this.commandId, () => {
```
rjrudin, line 18: Does this need to call `context.subscriptions.push` as well?

client/marklogicClient.ts
BillFarber: removed.

PR #149, opened 2023-03-30
DEVEXP-350: Standardize UX for JS and XQY

BillFarber:
Fire in the hole!

This PR is rather complex. Let me know if you'd like to review in a call.

Additionally, there are definitely improvements to be made. However, I think this is a major step in the right direction and we need to get these changes merged in.


rjrudin (changes requested):
The gist I get from the filenames is - we now have two config providers and two debug managers. Which tells me as a novice that there must be distinct differences between how debugging JS and XQY works. So that all seems fine to me.

I think the one thing to do is because this is both combining refactoring (moving a lot of code around) with some functional changes, change the commit comment to capture what those functional changes actually are. "Standardize" is useful but vague - a list of what is now standard between the two debuggers - from a user's perspective - would be helpful. Otherwise, I think the code looks good.

PR #154, opened 2023-03-31
DEVEXP-349: Run Highlighted Code

BillFarber:
This includes one new feature and some refactoring.

Run Highlighted Code : This allows a user "eval" only the selected code in the current editor tab. This works for all the eval functions (XQY, JS, SQL, SPARQL, Rows).

Refactoring : Completely "class-ified" the vscQueryDirectory and renamed the file and class to editorQueryEvaluator.


client/editorQueryEvaluator.ts
```diff
@@ -0,0 +1,221 @@
...
+    ROWS
+}
+
+export class EditorQueryEvaluator {
+    private static FOPTIONS = { tabSize: 2, insertSpaces: true };
+    private static FCOMMAND = 'vscode.executeFormatDocumentProvider';
+    private static SJS = 'sjs';
+    private static XQY = 'xqy';
+
+    private context: ExtensionContext;
```
rjrudin: I think `extensionContext` would be better now that we have two contexts flying around.

client/editorQueryEvaluator.ts
```diff
@@ -0,0 +1,221 @@
...
+            this.editorSqlQuery(client, actualQuery, resultsEditorTabIdentifier, editor, cfg);
+            break;
+        case EditorQueryType.ROWS:
+            this.editorRowsQuery(client, actualQuery, resultsEditorTabIdentifier, editor, rowsResponseFormat);
+            break;
+        }
+    }
+
+    private editorJSQuery(
+        db: ClientContext,
```
rjrudin: For new code, let's so `clientContext`, particularly as `db` is often used in the Node Client project to refer to an instance of the Node Client (which I don't like either; that happens in the Java Client project sometimes too, but I think "client" is a good name for a client object; it's of course not a database, it's a client to the database's REST API).

client/editorQueryEvaluator.ts
```diff
@@ -0,0 +1,221 @@
...
+    public constructor(context: ExtensionContext, provider: ClientResponseProvider) {
+        this.context = context;
+        this.provider = provider;
+    }
+
+    // Exported only for testing purposes
+    public static getQueryTextFromEditor(editor: TextEditor): string {
+        const selection = editor.selection;
+        const selectionText = editor.document.getText(selection);
+        let actualQuery: string = null;
```
rjrudin: I think we can just call this `query` - the use of `actual` makes me think that mlxprs took the user's query (whether it was the full editor text or only highlighted text) and then altered it in some fashion. But it's just a `query`, no need for any qualifier on it. And I think the method can be named `getQueryFromEditor`, and this method has the smarts to know whether the query is all of the text, or only the selected text.

client/editorQueryEvaluator.ts
```diff
@@ -0,0 +1,221 @@
...
+    public editorQuery(
+        editorQueryType: EditorQueryType, editor: TextEditor, rowsResponseFormat?: ml.RowsResponseFormat
+    ): void {
+        const actualQuery = EditorQueryEvaluator.getQueryTextFromEditor(editor);
+        const cfg: WorkspaceConfiguration = workspace.getConfiguration();
+        let language = EditorQueryEvaluator.SJS;
+        if (editorQueryType === EditorQueryType.XQY) {
+            language = EditorQueryEvaluator.XQY;
+        }
+        const client: ClientContext = cascadeOverrideClient(actualQuery, language, cfg, this.context.globalState);
```
rjrudin: Should use `clientContext` here, as I think `client` should refer only to the Node Client instance.

client/editorQueryEvaluator.ts
```diff
@@ -0,0 +1,221 @@
...
+        } else {
+            actualQuery = editor.document.getText();
+        }
+        return actualQuery;
+    }
+
+    public editorQuery(
+        editorQueryType: EditorQueryType, editor: TextEditor, rowsResponseFormat?: ml.RowsResponseFormat
+    ): void {
+        const actualQuery = EditorQueryEvaluator.getQueryTextFromEditor(editor);
```
rjrudin: I think this line shows why `query` is sufficient - it's using two terms for the same thing here = "query text" and "actual query". I think this is sufficient and simple:

    const query = EditorQueryEvaluator.getQueryFromEditor(editor);

rjrudin (changes requested):
Looks good, just had some renaming ideas for clarity.

client/editorQueryEvaluator.ts
BillFarber: I decided to go with dbClientContext.

PR #161, opened 2023-04-05
XQY formatting prototype

BillFarber:
XQY formatter from @quodatum/xqlint

### DIFF
--- a/client/extension.ts
+++ b/client/extension.ts
@@ -7,6 +7,7 @@ import { ClientContext } from './marklogicClient';
 import { cascadeOverrideClient } from './vscQueryParameterTools';
 import { ClientResponseProvider } from './clientResponseProvider';
 import { XmlFormattingEditProvider } from './xmlFormatting/Formatting';
+import { XqyFormattingEditProvider } from './xqyFormatting/XqyFormattingEditProvider';
 import { LanguageClient, LanguageClientOptions, ServerOptions, TransportKind } from 'vscode-languageclient/node';
 import { XqyDebugConfigurationProvider, XqyDebugAdapterDescriptorFactory } from './XQDebugger/xqyDebugConfigProvider';
 import { XqyDebugManager } from './XQDebugger/xqyDebugManager';
@@ -102,6 +103,12 @@ export function activate(context: vscode.ExtensionContext): void {
         vscode.languages.registerDocumentFormattingEditProvider(
             { scheme: 'mlquery', language: 'xsl' },
             new XmlFormattingEditProvider()
+        )
+    ]);
+    handleUnload(context, [
+        vscode.languages.registerDocumentFormattingEditProvider(
+            'xquery-ml',
+            new XqyFormattingEditProvider()
         )
     ]);
 
--- a/client/xqyFormatting/XqyFormatter.ts
+++ b/client/xqyFormatting/XqyFormatter.ts
@@ -1,0 +1,31 @@
+'use strict';
+
+import { window } from 'vscode';
+import { CodeFormatter, XQLint } from '@quodatum/xqlint';
+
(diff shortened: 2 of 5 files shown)


BillFarber:
This prototype depends on @quodatum/xqlint in NPM
https://socket.dev/npm/package/@quodatum/xqlint
https://www.npmjs.com/package/@quodatum/xqlint

That project is a fork of https://www.npmjs.com/package/xqlint
https://github.com/wcandillon/xqlint
https://www.npmjs.com/package/xqlint

The quodatum version works fine, but XQLint crashes when attempting to format.

I am a little concerned about using quodatum since the version number is 0.2.1, implying that it is incomplete. On the other hand, that doesn't seem to unusual in the NPM world.

I think that if we decide to use this package, we should be prepared to either:
1) Create our own fork which we maintain, or
2) Commit to assisting with the development and maintenance of quodatum.

The only other alternative would be to write our own parser and formatter from scratch. However, I think that's probably a VERY big job.

anu3990 (approved):

client/extension.ts
```diff
@@ -102,6 +103,12 @@ export function activate(context: vscode.ExtensionContext): void {
             new XmlFormattingEditProvider()
         )
     ]);
+    handleUnload(context, [
+        vscode.languages.registerDocumentFormattingEditProvider(
+            'xquery-ml',
```
rjrudin, line 110: I noticed other references to 'xquery-ml' - are we stuck with that? If not, I think renaming "ml" to "marklogic" would be good (as part of our goal to avoid people thinking "ml" means "machine learning"), but this is a very small thing and it's fine if we're stuck with it.

package.json
```diff
@@ -512,6 +513,7 @@
     "dependencies": {
         "rimraf": "4.3.1",
```
rjrudin: Shouldn't "rimraf" be a devDependency? I am assuming "dependencies" are included in the plugin, but we only need `rimraf` for some of the scripts used during development.

client/xqyFormatting/XqyFormatter.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export class XqyFormatter {
+    ast: unknown;
+
+    newLine: string;
+    indentPattern: string;
+    splitNamespaces: boolean;
+
+    formatXqy(xqy: string): string {
+        window.showInformationMessage('Formatting XQuery');
```
rjrudin, line 16: I think this should be removed - it doesn't appear for e.g. formatting JSON, and it stays there, forcing a user to have to remove it themselves.

client/xqyFormatting/XqyFormatter.ts
```diff
@@ -0,0 +1,31 @@
...
+        const linter = new XQLint(xqy, undefined, DEFAULT_OPTIONS);
+        const formatter = new CodeFormatter(linter.getAST());
+        this.ast = linter.getAST();
+        console.debug(linter.printAST());
```
rjrudin, line 20: I think all 3 of these debug statements should be removed - they seem useful for development (and thus can be easily added when developing), but not in a release.

rjrudin (commented):
My initial impression is that the xqy formatter produces some really undesirable results. Two easy examples:

```
(: Comment above a line :)
let $_ := "Hi"
let $_ := "Hey" (: Comment on the same line :)
let $_ := "Next line"
```

That gets formatted into this:

```
(: Comment above a line :)


let $_ := "Hi"
let $_ := "Hey"
(: Comment on the same line :) let $_ := "Next line"
```

I think this branch is worth keeping in place, but I'm thinking we either need to:

1. Wait for a "better" release of the xquery formatting library.
2. Or if it's possible, expose formatting options that quodatum supports so that the user has control over the formatting - though that would only help if it can avoid what seems like bad formatting behavior as shown above.

PR #164, opened 2023-04-11
DEVEXP-352: View results tab below editor tab

BillFarber:
1) This creates a Webview in the VSCode panel (the set of windows at the bottom of the application.

2) The Webview displays HTML.

3) Eval and query results are formatted for HTML and then displayed in the Webview.

### DIFF
--- a/client/editorQueryEvaluator.ts
+++ b/client/editorQueryEvaluator.ts
@@ -1,11 +1,14 @@
 'use strict';
-import { ExtensionContext, TextDocument, TextEdit, TextEditor, Uri, WorkspaceEdit,
-    commands, window, workspace, WorkspaceConfiguration } from 'vscode';
-import { cascadeOverrideClient } from './vscQueryParameterTools';
-import { ClientContext, sendJSQuery, sendSparql, sendXQuery, sendRows } from './marklogicClient';
-import { ClientResponseProvider } from './clientResponseProvider';
+import { XMLParser, XMLBuilder } from 'fast-xml-parser';
 import { contentType, RowsResponse } from 'marklogic';
 import * as ml from 'marklogic';
+import { ExtensionContext, TextDocument, TextEdit, TextEditor, Uri, WorkspaceEdit,
+    commands, window, workspace, WorkspaceConfiguration } from 'vscode';
+
+import { ClientResponseProvider, ErrorResultsObject } from './clientResponseProvider';
+import { ClientContext, sendJSQuery, sendSparql, sendXQuery, sendRows } from './marklogicClient';
+import { MlxprsWebViewProvider } from './mlxprsWebViewProvider';
+import { cascadeOverrideClient } from './vscQueryParameterTools';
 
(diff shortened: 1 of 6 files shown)


client/extension.ts
```diff
@@ -147,14 +149,23 @@ export function activate(context: vscode.ExtensionContext): void {
     XqyDebugManager.registerMlxprsStatusBarItem(mlxprsStatus);
     JsDebugManager.registerMlxprsStatusBarItem(mlxprsStatus);
     mlxprsStatus.requestUpdate();
+
+
+    const pmbWebViewProvider = new MlxprsWebViewProvider(context.extensionUri);
```
rjrudin: "pmb" = Bill M. Farber?

package.json
```diff
@@ -378,8 +396,14 @@
                 "type": "object",
                 "title": "MarkLogic Miscellaneous Configuration",
                 "properties": {
-                    "marklogic.sparqlContentType": {
+                    "marklogic.resultsRight": {
                         "order": 1,
+                        "type": "boolean",
+                        "default": false,
+                        "description": "Display query results in an editor tab to the right."
```
rjrudin: Is it guaranteed to be to the right? I'm wondering if it's safer to say "Display query results in a separate editor."

client/editorQueryEvaluator.ts
```diff
@@ -218,4 +281,78 @@ export class EditorQueryEvaluator {
...
+    private static requestMlxprsWebviewUpdateWithRecordArray(response: Record<string, unknown>[], contentType?: string): void {
+        if (this.mlxprsWebViewProvider) {
+            const stringResults =
+                (response as Record<string, unknown>[]).map(record => this.convertRecordToHtml(record, contentType)).join('\n');
+            this.mlxprsWebViewProvider.updateViewContent(stringResults);
+        }
+    }
+
+    private static convertRecordToHtml(record: Record<string, any>, contentType?: string): string {
+        if (record['format'] === 'xml') {
```
rjrudin, line 315: This all makes sense, but I'm wondering - what is a Record, and what are all the possible keys in it? Like why would "head" be populated but not "value"? Is this all VSCode stuff, or is it stuff we control?

rjrudin (commented):
Looks pretty good, and I think it would be fine to open a separate bug for the issue where the MLXPRS window doesn't work until a user clicks on it.

client/editorQueryEvaluator.ts
BillFarber: I don't have a complete understanding of Record. It is defined in a VSCode library, but it is used to define many of the responses from MarkLogic's Node client. 
These type definitions is an area that could probably benefit from more refactoring and documentation.

BillFarber:
> Looks pretty good, and I think it would be fine to open a separate bug for the issue where the MLXPRS window doesn't work until a user clicks on it.

Bug created: https://project.marklogic.com/jira/browse/DEVEXP-382

PR #165, opened 2023-04-11
DEVEXP-364: Attach to remote request won't work unless in a current source file

BillFarber:


client/JSDebugger/jsDebugConfigProvider.ts
```diff
@@ -24,13 +24,15 @@ export class JsDebugConfigurationProvider implements vscode.DebugConfigurationPr
...
         }
 
-        // use active editor window as ad-hoc query if `program` not defined explicitly
-        config.program = config.program || config.path;
-        if (!config.program) {
-            const doc: vscode.TextDocument = vscode.window.activeTextEditor.document;
-            config.program = doc.fileName;
-            config.scheme = doc.uri.scheme;
-            config.queryText = doc.getText();
+        // if 'launch', then use active editor window as ad-hoc query if `program` not defined explicitly
```
rjrudin: I'm not quite getting this yet - it seems like the comment should explain what's going on as something like "If the (user did whatever would result on config.request = launch), then it's possible that (whatever config.program refers to) may be null, in which case set it to (whatever config.path refers to)."

I guess what I'm really asking is - as a new VSCode developer, I don't know what "config.request", "config.program", and "config.path" refer to.

PR #169, opened 2023-04-13
DEVEXP-382: Eval results are not displayed in Webview unless Webview has been displayed previously.

BillFarber:
I figured out a hack to show the results in the webview the first time it is run.

### DIFF
--- a/client/editorQueryEvaluator.ts
+++ b/client/editorQueryEvaluator.ts
@@ -2,8 +2,10 @@
 import { XMLParser, XMLBuilder } from 'fast-xml-parser';
 import { contentType, RowsResponse } from 'marklogic';
 import * as ml from 'marklogic';
-import { ExtensionContext, TextDocument, TextEdit, TextEditor, Uri, WorkspaceEdit,
-    commands, window, workspace, WorkspaceConfiguration } from 'vscode';
+import {
+    ExtensionContext, TextDocument, TextEdit, TextEditor, Uri,
+    WorkspaceEdit, commands, window, workspace, WorkspaceConfiguration
+} from 'vscode';
 
 import { ClientResponseProvider, ErrorResultsObject } from './clientResponseProvider';
 import { ClientContext, sendJSQuery, sendSparql, sendXQuery, sendRows } from './marklogicClient';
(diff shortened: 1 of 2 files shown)


client/editorQueryEvaluator.ts
```diff
@@ -2,8 +2,10 @@
 import { XMLParser, XMLBuilder } from 'fast-xml-parser';
 import { contentType, RowsResponse } from 'marklogic';
 import * as ml from 'marklogic';
-import { ExtensionContext, TextDocument, TextEdit, TextEditor, Uri, WorkspaceEdit,
-    commands, window, workspace, WorkspaceConfiguration } from 'vscode';
+import {
```
rjrudin, line 5: Just curious - is there a VSCode'ing convention for formatting multiple imports like this? I can think of the following:

1. All on one line
2. As many as can fit per line
3. One per line

I guess that's a generic JS/Typescript question too. I ask because Java does one per line, which I've always liked. 

I do respect the alphabetizing going on here.

client/editorQueryEvaluator.ts
```diff
@@ -283,58 +285,61 @@ export class EditorQueryEvaluator {
     }
 
     private static requestMlxprsWebviewUpdateWithErrorResultsObject(response: ErrorResultsObject): void {
-        if (this.mlxprsWebViewProvider) {
-            const stringResults = this.convertTextResponseToHtml(JSON.stringify(response, null, 2));
-            this.mlxprsWebViewProvider.updateViewContent(stringResults);
+        if (EditorQueryEvaluator.mlxprsWebViewProvider) {
+            commands.executeCommand('mlxprs.ResultsView.focus');
```
rjrudin: This is called 3 times, and I think it could use an explanation since it's the magic line here, so I think it's worth a separate method - e.g. `focusOnResultsView` - where the method comments explain why this is necessary, perhaps including a link to one of those URLs you sent that identified the problem. In fact, a link to the Github ticket where VSCode users are asking for a "reveal" method would be perfect.

client/editorQueryEvaluator.ts
BillFarber: From what I can tell, there are no strict conventions at this time, but I haven't searched real hard yet.

I've been trying to gradually adjust the imports as I refactor things. My strategy is basically to have 2 groups. The first for 3rd party imports and the second for imports from this project. Those groups are then sorted alphabetically by the package name and individual components are sorted there as well. 

I just discovered there is an ESLint setting for this, "import/order". Perhaps I should try this out. https://dev.to/diballesteros/how-to-quickly-configure-eslint-for-import-sorting-2h73

rjrudin (approved):
Working for me

PR #172, opened 2023-04-21
DEVEXP-401: Replace the "request-promise" package in marklogicClient.ts

BillFarber:
Replaced the "request-promise" package with the "needle" package for the single http request in this source file.
Also did some linting and formatting.

### DIFF
--- a/client/marklogicClient.ts
+++ b/client/marklogicClient.ts
@@ -1,9 +1,11 @@
 'use strict';
 
-import * as request from 'request-promise';
-import * as ml from 'marklogic';
 import * as fs from 'fs';
+import * as ml from 'marklogic';
+import needle = require('needle');
+
 import { MlxprsStatus } from './mlxprsStatus';
+
 
 export const MLDBCLIENT = 'mldbClient';
 export const MLSETTINGSFLAG = /mlxprs:settings/;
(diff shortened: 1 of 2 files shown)


client/marklogicClient.ts
```diff
@@ -340,13 +349,16 @@ export async function getFilteredListOfJsAppServers(
...
+export async function getAppServerListForJs(
+    dbClientContext: ClientContext, serverListQuery: string
+): Promise<MlxprsQuickPickItem[]> {
     return sendXQuery(dbClientContext, serverListQuery)
         .result(
             (appServers: Record<string, ServerQueryResponse>) => {
                 return dbClientContext.buildAppServerChoicesFromServerList(appServers);
             },
             err => {
+                console.error('Error: ' + JSON.stringify(err));
```
rjrudin, line 361: Is this adding at least something so that the error isn't silently ignored?

client/marklogicClient.ts
```diff
@@ -358,24 +370,26 @@ export async function filterJsServerByConnectedStatus(
...
-            },
-            auth: {
-                user: dbClientContext.params.user,
-                pass: dbClientContext.params.pwd,
-                'sendImmediately': false
             }
         };
+        const options = {
+            headers,
+            username: dbClientContext.params.user,
```
rjrudin, line 381: Just confirming, mlxprs only supports digest/basic auth today, right? This of course won't work with MarkLogic Cloud or any other non-username/password auth strategy that MarkLogic supports. I think that's fine for the 3.x timeframe as it's likely always been the case for mlxprs. But at some point, we'd want mlxprs to support every auth strategy. 

That makes me wonder - instead of trying to upgrade `request` to a separate HTTP client - which ultimately requires supporting every auth type possible - is the better longterm approach here to figure out how to provide a backdoor in the Node Client for invoking any URL in MarkLogic? That would also need to include reusing the recently-added `basePath` support.

PR #174, opened 2023-04-22
DEVEXP-364: Attach to remote request won't work unless in a current source file

BillFarber:
Added some comments to clarify the code.

I missed the "attach to remote xquery request" the first time through this story.

### DIFF
--- a/client/XQDebugger/xqyDebugConfigProvider.ts
+++ b/client/XQDebugger/xqyDebugConfigProvider.ts
@@ -106,11 +106,17 @@ export class XqyDebugConfigurationProvider implements DebugConfigurationProvider
             }
         }
 
-        if (!config.program) {
-            config.program = window.activeTextEditor.document.fileName;
-            config.query = window.activeTextEditor.document.getText();
-        } else {
-            config.query = readFileSync(config.program).toString();
+        // Need to tell VSCode what parameters to use for "launching" a debug session based on a configuration in launch.json
+        // Note that this information is unnecessary for "attaching" to a remote request for debugging.
+        if (config.request === 'launch') {
+            if (!config.program) {
+                // If a program isn't specified by the launch configuration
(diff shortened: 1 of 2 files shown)


client/XQDebugger/xqyDebugConfigProvider.ts
```diff
@@ -106,11 +106,17 @@ export class XqyDebugConfigurationProvider implements DebugConfigurationProvider
...
 
-        if (!config.program) {
-            config.program = window.activeTextEditor.document.fileName;
-            config.query = window.activeTextEditor.document.getText();
-        } else {
-            config.query = readFileSync(config.program).toString();
+        // Need to tell VSCode what parameters to use for "launching" a debug session based on a configuration in launch.json
+        // Note that this information is unnecessary for "attaching" to a remote request for debugging.
+        if (config.request === 'launch') {
+            if (!config.program) {
```
rjrudin, line 112: Why does this conditional differ based on whether it's XQuery or JS? JS doesn't have an `else` condition, for example. I was initially thinking that it would be a helper function with common logic for both - i.e. "If no program is specified, then use the current editor".

BillFarber: My guess is that the "else" clause was added to the xqy version when somebody wanted to debug a specific module without having that module active (maybe they were editing a library module).
The same requirement probably just never popped up in JS.
I'm happy to add this to the JS version if you think I should.

rjrudin (approved):

client/XQDebugger/xqyDebugConfigProvider.ts
BillFarber: Added a new story to clean this up a bit and to make it clearer to the user.
https://project.marklogic.com/jira/browse/DEVEXP-403

PR #175, opened 2023-04-24
Fix JS use of "program" and update README

BillFarber:


README.md
```diff
@@ -201,7 +201,8 @@ Example 'launch' type configuration items for JavaScript and XQuery:
 ```
 
 By default, launch mode will launch the currently opened file for debugging.
-An optional "path" (for JavaScript) or "program" (for XQuery) can be provided to specify another file for debugging.
+An optional "program" property can be provided in a launch.json task configuration to specify another file for debugging.
```
rjrudin: For consistency with other areas in the README, use Markdown tick marks instead of double quotes. Should do that around `launch.json` and `program` and `path`. It's not 100% consistent that way - I doubt we have few README/CONTRIB files where we are 100% consistent with tick marks - but it should be done here since they're used consistently in the section right above this.

rjrudin (changes requested):

PR #177, opened 2023-05-02
Implement additional SPARQL return types

bshalke:


server/package-lock.json
rjrudin, line 1: If this only changed because of the whitespace change in `server/package.json`, can you undo the changes to both files so that the overall changeset is smaller?

client/editorQueryEvaluator.ts
```diff
@@ -210,7 +211,8 @@ export class EditorQueryEvaluator {
         resultsEditorTabIdentifier: Uri,
         editor: TextEditor
     ): void {
-        const contentType: contentType = workspace.getConfiguration().get('marklogic.sparqlContentType');
+        const queryType: ml.GraphsQueryType = getSparqlQueryForm(sparqlQuery);
```
rjrudin: Because this command is hitting "/v1/graphs/sparql", and because the command is called "Eval SPARQL", we should consistently refer to this as "sparqlQuery" and not introduce "graphsQuery". The latter could be misinterpreted as referring to the "/v1/graphs" endpoint too which has a different set of inputs and output types.

So all occurrences of "GraphsQuery" should be renamed to "SparqlQuery" for consistency - e.g. `SparqlQueryType`.

client/editorQueryEvaluator.ts
```diff
@@ -210,7 +211,8 @@ export class EditorQueryEvaluator {
...
+        const contentType: ml.GraphsResponseFormat = getSparqlResponseType(queryType);
```
rjrudin: Based on the docs - https://docs.marklogic.com/guide/semantics/REST#id_54258 - call this `SparqlOutputFormat`. The docs do use the term "Response Output Format" in the section above that, but that's a little redundant, and "output format" is well understood as it's used in other projects too for the same thing.

client/sparqlTools.ts
rjrudin, line 1: I'd like to avoid the term "Tools" as that's not widely used in any of our projects. I saw there's a `vscQueryParameterTools.ts` module, though the functions in there seem like "util" methods. And mlxprs does use "util" for the `RangeUtil.ts` module - though the modules in the client/xmlFormatting directory unfortunately use CamelCase for no discernible reason. 

@BillFarber  Any ideas on how to name a module that is geared around a particular command and a particular REST endpoint? I would lean towards `sparqlQueryManager.ts`.

rjrudin (changes requested):

client/sparqlTools.ts
BillFarber: sparqlQueryHelper.ts ?

We could probably use a function in this file that is capable of formatting any SPARQL result type for display in an HTML view.

client/editorQueryEvaluator.ts
bshalke: I aligned the type naming to be consistent with the existing graphs interface and rows format (RowsResponseFormat).  Should we change those as well?

rjrudin: Hmm - "ResponseFormat" isn't great (the "format" of the response is something that conforms to the HTTP specification), but good point - let's stick with that for consistency, so yeah, `SparqlResponseFormat`. But I think we should definitely use "sparql" instead of "graphs" since this is specific to the "/v1/graphs/sparql" endpoint and not the "/v1/graphs" endpoint.

bshalke: I went in and renamed the response types but I can't rename the interface because it's defined by the Node API:
https://docs.marklogic.com/jsdoc/graphs.html#sparql

rjrudin (commented):
This looks pretty good to me, but I defer to @BillFarber . I am thinking that it should be possible to add tests for this? Would it help if we had an ml-gradle-based test app that loads up some triples so that it's easy to write a test that uses mlxprs to connect to that app?

BillFarber (approved):

BillFarber:
> This looks pretty good to me, but I defer to @BillFarber . I am thinking that it should be possible to add tests for this? Would it help if we had an ml-gradle-based test app that loads up some triples so that it's easy to write a test that uses mlxprs to connect to that app?

I need to refactor/create a better MarkLogic test application for integration tests.
I'd like to merge this PR and then I will work on the refactor.
After that, @bshalke can add a couple of tests.

PR #178, opened 2023-05-02
DEVEXP-416: Bug fix when CA file doesn't exist

BillFarber:
We verify the existing of the specified CA file before passing control back to VSCode.

### DIFF
--- a/client/XQDebugger/xqyDebugConfigProvider.ts
+++ b/client/XQDebugger/xqyDebugConfigProvider.ts
@@ -14,14 +14,16 @@
  * limitations under the License.
  */
 
+import { readFileSync } from 'fs';
 import {
     DebugConfiguration, DebugConfigurationProvider, WorkspaceFolder, CancellationToken, ProviderResult,
     window, DebugAdapterDescriptorFactory, DebugAdapterExecutable, DebugAdapterDescriptor,
     DebugSession, QuickPickItem, QuickPickOptions, WorkspaceConfiguration, workspace
 } from 'vscode';
-import { readFileSync } from 'fs';
-import { XqyDebugManager, DebugStatusQueryResponse } from './xqyDebugManager';
+
+import { ErrorReporter, MlxprsError } from '../errorReporter';
 import { MlClientParameters } from '../marklogicClient';
+import { XqyDebugManager, DebugStatusQueryResponse } from './xqyDebugManager';
 
 
 export class XqyDebugConfiguration implements DebugConfiguration {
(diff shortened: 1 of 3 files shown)


client/XQDebugger/xqyDebugConfigProvider.ts
```diff
@@ -64,6 +66,23 @@ export class XqyDebugConfigurationProvider implements DebugConfigurationProvider
         });
         config.clientParams = clientParams;
 
+        if (clientParams.pathToCa) {
+            try {
+                const caFileContents = await readFileSync(clientParams.pathToCa);
+                if (!caFileContents || caFileContents.length === 0) {
+                    throw new Error('CA file contents are empty');
```
rjrudin: I think it'd be nice to add the value of `pathToCa` in both places; that helps the user understand what they messed up.

Also, for consistency, we should use either the label in the settings - "Path to Ca" - or the description in the settings - "path to certificate authority". "Path to Ca" of course is a little wonky (I think because VSCode requires CamelCase in those labels??). So I'd go with the latter - e.g. `"Path to certificate authority is an empty file; path: " + clientParams.pathToCa`.

client/XQDebugger/xqyDebugConfigProvider.ts
```diff
@@ -64,6 +66,23 @@ export class XqyDebugConfigurationProvider implements DebugConfigurationProvider
...
+                }
+            } catch (error) {
+                const mlxprsError: MlxprsError = {
+                    reportedMessage: error.message,
+                    stack: error.stack,
+                    popupMessage: 'CA file is unreadble or the contents are empty'
```
rjrudin: Same thing here - I'm thinking `"Path to certificate authority is an unreadable file or the file is empty"` and make sure the value of `pathToCa` is shown somewhere.

client/errorReporter.ts
```diff
@@ -0,0 +1,34 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { window } from 'vscode';
+
+export interface MlxprsError {
+    popupMessage: string;
+    reportedMessage: string;      // module URI on MarkLogic
```
rjrudin: These comments are a little surprising - `reportedMessage` above is an error message, I don't see a module URI being present.

BillFarber: My "copy & paste" oversight.

PR #180, opened 2023-05-03
Put all VSCode configuration settings requests in one place

BillFarber:
There are more than 20 places where the extension asks VSCode for settings.

I'm hoping that setting up this singleton will allow me to more easily override specific configuration settings for tests.

Other benefits:
1) All the interactions with VSCode settings are in one place
2) Possibly some caching benefits

### DIFF
--- a/client/configurationManager.ts
+++ b/client/configurationManager.ts
@@ -1,0 +1,97 @@
+/*
+ * Copyright (c) 2023 MarkLogic Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


client/configurationManager.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export class ConfigurationManager {
+    private static vscodeConfiguration: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration();
+
+    static getVscodeConfiguration(): vscode.WorkspaceConfiguration {
+        return ConfigurationManager.vscodeConfiguration;
+    }
+
+    static handleUpdateConfigurationEvent(): void {
+        vscode.window.showErrorMessage('handleUpdateConfigurationEvent');
```
rjrudin: Just curious, what does this line actually do? It seems odds to show an error message when a user changes config, is it doing something else?

BillFarber: This was just me verifying that when I changed a setting, it would immediately fire and the singleton would be updated.
If we decide to actually merge this in, this line will be removed.

rjrudin (approved):
I would even go further and consider adding well-named functions to the config manager class - e.g. "getPort", "getPathToCa". Only if it helps though.

PR #181, opened 2023-05-04
DEVEXP-419: Show error when "Disconnect" can't get the info

BillFarber:
Displays standardized error messages now when using one of the "Disconnect" commands fails to connect to the MarkLogic server.

Also did some import organizing.


rjrudin (approved):
I verified that Disconnect with bad settings gives me a nice error

PR #183, opened 2023-05-11
DEVEXP-420: Launching Eval fails with self-signed cert

BillFarber:
This one requires manual testing.
The primary changes for this story:
* extension - register the new custom message handler
* xqyDebug - stops the request when launching with ML fails.
* xqyDebugManager - added a handler for custom events from the Debug Adapter, including the new one generated when launching with ML fails.
* xqyRuntime - when launching the query with ML fails, creates a new standard MLXPRS error object and returns that object to the debug client via a custom event.

I attempted to write an integration test for it, but was unable to get the dynamic Debug Client to work properly with the dynamic XQY Debug Adapter.
However, during that attempt, I did some refactoring to the integration test directory that will help with future efforts at XQY debugging integration testing.

Also:
Added some debug-level logging for interactions with the Debug Adapters.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -67,13 +67,15 @@ It is recommended to use VSCode as the editor for this package, as it can self-l
     * Note that at this time, the "Attach" commands do not work when debugging the extension from within VSCode. In order to test the "Attach" commands, you will need to build the artifact (.vsix) and use that extension with a different project, and then test manually.
 * Please see the README.md file for information on configuring and working in the test environment
 
-## Debugging the Debugger
-This can be just a little tricky. The best way I've found to test/debug the debugger from withing VSCode is to use the
-"Extension + SJS Server" or "Extension + XQY Server" launch configuration. Using one of these configurations, VSCode will start a new VSCode window where you can have a different project open for testing the debugger. When you use one
-of the "attach" configurations from the second window to attach to a process, the second window also connects to the 
-original VSCode window for debugging purposes. At that point, you have 2 VSCode windows, both attached to a process
-for debugging - the first VSCode is attached to the second VSCode, and the second VSCode is attached to the running copy
-of its project.
+## Debugging the Debugger/Evaluate tasks
+This can be just a little tricky. The preferred mechanism to test/debug the debugger from within VSCode is to use the "Extension + SJS Server" or "Extension + XQY Server" launch configuration. Using one of these configurations, VSCode will start a new VSCode window where you can have a different project open for testing the debugger. VSCode also starts an instance of the Debug Adapter. When you use one of the "attach" or "evaluate" launch.json configurations from the second window to attach to a process, the second window connects to the Debug Adapter in the original VSCode window for debugging purposes. At that point, you have effectively have 3 debug sessions going on:
+1. The debug/launch session in the second window. This, presumably, is the debugger you are working on.
+2. The debug session for the debugger you're working on. Shown as "Launch Extension (preprod)" in the Call Stack view.
+    * This will show the code from the mlxprs project, other than the four source files mentioned in #3.
+3. The debug session for the Debug Adapter. Shown as "Launch <XQY|SJS> Debug Adapter Server" in the Call Stack view.
+    * This will show the code from xqyDebug, xqyRuntime, mlDebug, and mlRuntime
+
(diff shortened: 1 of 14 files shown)


client/XQDebugger/xqyDebug.ts
```diff
@@ -177,15 +178,22 @@ export class XqyDebugSession extends LoggingDebugSession {
 
     protected async launchRequest(response: DebugProtocol.LaunchResponse, args: LaunchRequestArguments): Promise<void> {
         logger.setup(Logger.LogLevel.Stop, false);
+        logger.setup(Logger.LogLevel.Verbose, false);
+        console.debug(JSON.stringify(args));
```
rjrudin: Should this be left in? Could args have a password value in it?

CONTRIBUTING.md
```diff
@@ -67,13 +67,15 @@ It is recommended to use VSCode as the editor for this package, as it can self-l
...
-of the "attach" configurations from the second window to attach to a process, the second window also connects to the 
-original VSCode window for debugging purposes. At that point, you have 2 VSCode windows, both attached to a process
-for debugging - the first VSCode is attached to the second VSCode, and the second VSCode is attached to the running copy
-of its project.
+## Debugging the Debugger/Evaluate tasks
+This can be just a little tricky. The best way I've found to test/debug the debugger from within VSCode is to use the "Extension + SJS Server" or "Extension + XQY Server" launch configuration. Using one of these configurations, VSCode will start a new VSCode window where you can have a different project open for testing the debugger. VSCode also starts an instance of the Debug Adapter. When you use one of the "attach" or "evaluate" launch.json configurations from the second window to attach to a process, the second window connects to the Debug Adapter in the original VSCode window for debugging purposes. At that point, you have effectively have 3 debug sessions going on:
```
rjrudin: We generally want to avoid first person here - just say something like "The preferred mechanism to test/debug the debugger...".

client/test/integration/index.ts
```diff
@@ -39,8 +39,8 @@ export async function run(): Promise<void> {
 
     mocha.options.color = true;
     mocha.rootHooks({
-        beforeEach: () => {
-            globalThis.integrationTestHelper.setupEachTest();
+        beforeEach: async () => {
```
rjrudin, line 42: This raises the question of - why is async/await needed here but not for afterEach? And why is it being included - what problem is being fixed? I think a comment in the code is the right way to describe why.

client/test/integration/markLogicIntegrationTestHelper.ts
```diff
@@ -44,12 +45,13 @@ export class IntegrationTestHelper {
...
     private password = String(process.env.ML_PASSWORD || 'admin');
     private modulesDB = String(process.env.ML_MODULESDB || this.modulesDatabase);
-    private pathToCa = null;
+    private pathToCa = '';
     private ssl = false;
     private rejectUnauthorized = false;
 
     public config = null;
-    public debugClient: DebugClient = null;
+    public jsDebugClient: DebugClient = null;
```
rjrudin, line 53: Nice.

client/test/integration/sjsAdapter.test.ts
```diff
@@ -33,52 +33,70 @@ suite('Testing sjs/xqy boundary in eval/invoke', async () => {
...
-        await debugClient.stepInRequest({ threadId: 1 });
-        return debugClient.assertStoppedLocation('step', { path: globalConfig.program, line: 6 });
+        await jsDebugClient.setBreakpointsRequest({ source: { path: globalConfig.program }, breakpoints: [{ line: 4 }] });
+        await jsDebugClient.continueRequest({ threadId: 1 });
+        await jsDebugClient.assertStoppedLocation('breakpoint', { path: globalConfig.program, line: 4 });
+        await jsDebugClient.stepInRequest({ threadId: 1 });
+        return jsDebugClient.assertStoppedLocation('step', { path: globalConfig.program, line: 6 });
     }).timeout(15000);
 
+    // test('launch xqy', async () => {
```
rjrudin: Is this going to be needed again soon?

rjrudin (approved):
All LGTM, I just had a couple small questions.

client/test/integration/index.ts
BillFarber: Good point. The teardown is asynchronous also, so I updated that code to use async/await.

client/XQDebugger/xqyDebug.ts
BillFarber: Good catch. Removed.

client/test/integration/sjsAdapter.test.ts
BillFarber: I guess not.
Dammit, I hate giving up.

PR #186, opened 2023-05-16
Adding ESLint config to the test-app

BillFarber:


test-app/.editorconfig
```diff
@@ -0,0 +1,22 @@
...
+[*]
+indent_style = space
+indent_size = 2
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.js]
+indent_size = 4
```
rjrudin, line 13: I think we should stick with 2 - that's what we have in Koop and GDS, and while node-client-api doesn't have an .editorconfig (it should!), it looks like 2 is consistently used in the files under ./lib in node-client-api. 

Same goes for .json and .ts below. Basically, .java is the oddball for us with an indent_size of 4, but that's long been a standard for Java.

rjrudin (changes requested):

BillFarber:
Note that I just copied this file from the root of mlxprs, so if we change
it here, I would want to change it there as well.
That's fine, but it would reformat the entire project.
Still want me to change it?

test-app/.editorconfig
BillFarber: Note that I just copied this file from the root of mlxprs, so if we change it here, I would want to change it there as well.
That's fine, but it would reformat the entire project.
Still want me to change it?

rjrudin: Got it, that makes sense, so no change.

rjrudin (approved):

test-app/.editorconfig
BillFarber: Since it's like this in the root MLXPRS project, I'm pushing this issue for later.

PR #188, opened 2023-05-18
DEVEXP-424: Show "Evaluate Current JavaScript Module" errors in a popup

BillFarber:
Also added a clone of the "Manage" app-server with SSL turned on to facilitate testing.

If you attempt to run the automated tests locally, note the "note" that was added to the CONTRIB file.


rjrudin (approved):
Looks good, now that we know how to get the SSL cert template working.

PR #189, opened 2023-05-18
A couple of small updates

BillFarber:


rjrudin (approved):
Could also bump to ml-gradle 4.5.2 if you'd like.

PR #191, opened 2023-05-22
DEVEXP-426: Show "Connect" and "Disconnect" errors in a popup

BillFarber:


client/test/integration/markLogicIntegrationTestHelper.ts
```diff
@@ -98,6 +98,21 @@ export class IntegrationTestHelper {
             rejectUnauthorized: true
         })
     );
+    readonly mlClientWithBadPort = new ClientContext(
+        new MlClientParameters({
```
rjrudin, line 102: I think a good refactor (for later, not in this ticket) would be to minimize the number of times where all these properties get spelled out. That makes it harder to add a new auth property in the future - like when we add support for ML Cloud. 

At least in this case, an e.g. `newParams` function in this module would be helpful for creating an `MlClientParameters` based on all of the values in this module, and that function could allow for any other params to be passed in as well. Would also make it easy to see what's being overridden as well - for example, we'd ideally only see `port: 9999` here.

rjrudin (approved):

client/test/integration/markLogicIntegrationTestHelper.ts
BillFarber: I like that. Good idea.

PR #194, opened 2023-06-05
Replace the requests package - part 1

BillFarber:
No new tests since existing tests exercised the re-factored code. However, the refactor changes some error reporting, so some tests have changed a bit.

This is to be merged with the develop branch and will be the first PR for 3.8.0.
I have already created and pushed a "release/3.7.0" branch.

### DIFF
--- a/client/JSDebugger/mlRuntime.ts
+++ b/client/JSDebugger/mlRuntime.ts
@@ -15,11 +15,11 @@
  */
 
 import { EventEmitter } from 'events';
+import * as fs from 'fs';
+import * as ml from 'marklogic';
+
 import { ClientContext, MlClientParameters } from '../marklogicClient';
 import { ModuleContentGetter } from '../moduleContentGetter';
-import * as request from 'request-promise';
-import * as fs from 'fs';
-import * as querystring from 'querystring';
 
 const DEFAULT_MANAGE_PORT = 8002;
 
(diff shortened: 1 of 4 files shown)


client/JSDebugger/mlRuntime.ts
```diff
@@ -245,70 +247,104 @@ export class MLRuntime extends EventEmitter {
...
+                (operation: ml.RequestOperation) => {
+                    operation.requestBody = body;
+                })
+                .result((result: object) => {
+                    let response = '';
+                    if (result) {
+                        response = JSON.stringify(result);
+                    }
+                    resolve(response);
+                }).catch(error => {
```
rjrudin: Very small thing - this reads very nicely, but I'm thinking the ".catch" should be on its own line so that the ".result" and ".catch" really stand out. That might not be the case in much of the codebase, but in general, I like that for when a bunch of functions are chained together as it makes each one stand out.

client/JSDebugger/mlRuntime.ts
```diff
@@ -245,70 +247,104 @@ export class MLRuntime extends EventEmitter {
...
-            }
-        };
-        if (queryString) {
-            options['qs'] = queryString;
+    private _sendMLdebugRequestGET(module: string, queryStringObject?: Record<string, unknown>): Promise<string> {
+        const manageClientParams: MlClientParameters = JSON.parse(JSON.stringify(this.dbClientContext.params));
+        manageClientParams.port = this.managePort;
+        manageClientParams.contentDb = null;
+        manageClientParams.modulesDb = null;
+        const manageClient = new ClientContext(manageClientParams);
```
rjrudin: These 5 lines are repeated above - perhaps a `newManageClient` function is warranted here?

rjrudin (approved):
LGTM, I just had a couple small suggestions for formatting / further refactoring.

PR #197, opened 2023-06-07
Get the test-app setup with marklogic-unit-test

BillFarber:


test-app/build.gradle
```diff
@@ -1,3 +1,12 @@
+buildscript {
+  repositories {
+    mavenCentral()
+  }
+  dependencies {
```
rjrudin: You only need this if you plan to run `mlUnitTest`.

rjrudin (approved):

PR #199, opened 2023-06-10
Add more demo capabilities to the test-app

BillFarber:
Samples for debugging XQY & JS marklogic-unit-tests
Samples for debugging internal Modules


test-app/src/main/ml-modules/root/javascript/callOptic.sjs
```diff
@@ -0,0 +1,15 @@
+const op = require('/MarkLogic/optic');
+
+/*
```
rjrudin: This seems like something that would be more likely to be read in the CONTRIB guide, which would have a section on how to test stepping into built-in MarkLogic code. And it wouldn't need to be duplicated in the callOptic.xqy file either.

rjrudin (approved):

PR #205, opened 2023-06-13
Merging the rebased & merged develop into origin develop

BillFarber:
This the result of merging the single (so far) 3.8.0 commit with the result of merging release/3.7.0 into master.

This puts the CHANGELOG updates in the wrong order and the version in server/package.json will need to be updated.


rjrudin:
Merge #204 first and then rebase develop off master.

PR #209, opened 2023-06-14
Removing "requests" from marklogicClient

BillFarber:
Also, added a test for the function where "requests" was being used.

### DIFF
--- a/client/marklogicClient.ts
+++ b/client/marklogicClient.ts
@@ -18,7 +18,6 @@
 
 import * as fs from 'fs';
 import * as ml from 'marklogic';
-import * as request from 'request-promise';
 
 import { MlxprsStatus } from './mlxprsStatus';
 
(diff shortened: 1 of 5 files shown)


client/marklogicClient.ts
```diff
@@ -363,30 +363,27 @@ export async function filterJsServerByConnectedStatus(
...
-        }
-        options['rejectUnauthorized'] = dbClientContext.params.rejectUnauthorized;
-        const connectedRequest = request.get(url, options)
-            .then((response) => {
-                if (response === requiredResponse) {
+
+    const manageClientParams: MlClientParameters = JSON.parse(JSON.stringify(dbClientContext.params));
+    manageClientParams.port = managePort;
+    manageClientParams.contentDb = null;
+    manageClientParams.modulesDb = null;
```
rjrudin: This is the same as `newMarkLogicManageClient` in `mlRuntime.ts` - perhaps that function should instead be in here so that it can be reused both here and in `mlRuntime.ts`?

BillFarber: I thought about it. Decided it wasn't worth it. But based on your comment, I've gone ahead and refactored it.

rjrudin (approved):

PR #211, opened 2023-06-23
Finished replace the requests library

BillFarber:
Replaced request in jsDebugManager.getAvailableRequests
Replaced request in jsDebugManager.resolveDatabasetoId
Replaced request in jsDebugManager.getRequestInfo

Added tests for each of those functions.


client/test/integration/markLogicIntegrationTestHelper.ts
```diff
@@ -174,6 +176,14 @@ export class IntegrationTestHelper {
             await JsDebugManager.disconnectFromNamedJsDebugServer(globalThis.integrationTestHelper.attachServerName);
             globalThis.integrationTestHelper.attachedToServer = false;
         }
+        // Some tests can leave the MarkLogic server in a state that causes future tests to fail.
+        // The only reliable method for resolving this state is to restart the MarkLogic server.
+        if (globalThis.integrationTestHelper.restartServer) {
+            await this.restartMarkLogic();
+            globalThis.integrationTestHelper.restartServer = false;
+            await wait(5000);
```
rjrudin: Is the hope that 5s is long enough for MarkLogic to restart? The /admin/v1/timestamp endpoint can be used in a loop instead to reliably wait for MarkLogic to restart, that's what ml-app-deployer does.

client/test/integration/index.ts
```diff
@@ -39,6 +39,7 @@ export async function run(): Promise<void> {
     // mocha.reporter('xunit', { output: resultsFileName });
 
     mocha.options.color = true;
+    mocha.timeout(15000);
```
rjrudin: Why 15s? I think it's good to comment all wait/timeout values like this so a developer six months from now understands why this particular value was chosen and not e.g. 10s or 20s.

client/JSDebugger/jsDebugManager.ts
```diff
@@ -220,57 +225,73 @@ export class JsDebugManager {
...
+                    resolve(null);
+                });
+        });
     }
 
-    public static async getRequestInfo(username: string, password: string, requestId: string, debugServerName: string, hostname: string,
-        ssl?: boolean, ca?: Buffer, rejectUnauthorized = true, managePort = ClientContext.DEFAULT_MANAGE_PORT
-    ): Promise<string> {
+    public static async getRequestInfo(requestId: string, debugServerName: string): Promise<string> {
+        const mlManageClientParameters = buildNewManageClientParametersFromConfig();
```
rjrudin: This block is in a few places - how about:

`const manageClient = newManageClient()`? 

Also, I think you can get away with `newManageConfig()` instead of `buildNewManageClientParametersFromConfig()`, as "manageConfig" is widely used in ml-app-deployer to represent the same concept.

rjrudin (approved):

PR #213, opened 2023-07-07
DEVEXP-393 - Run marklogic-unit-test module

BillFarber:
Adds the feature to run a single ML Unit Test - based on the current editor tab.

Added a new configure option to specify the port that the unit tests are running on.

Update README - including unrelated tweaks.

Moved a couple helper functions that are only used for the MLXPRS webview to the webview provider.

Includes a little refactoring of MarkLogic client generation in the TestHelper file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,8 +12,9 @@ _Develop, run, and debug code for MarkLogic in the popular VS Code IDE_
 
 * Syntax highlighting and IntelliSense for MarkLogic Server-Side JavaScript and XQuery
 * Interactive debugging of JavaScript and XQuery running in MarkLogic, including attaching to in-flight requests and inspecting live variables
-* Real-time query evaluation of JavaScript, XQuery, SQL, and SPARQL against a live Data Hub Service or MarkLogic instance
+* Real-time query evaluation of JavaScript, XQuery, SQL, SPARQL, and Optic against a MarkLogic instance
 * View modules (read-only) in the editor
+* Run [marklogic-unit-test module](https://github.com/marklogic-community/marklogic-unit-test)
 
 _JavaScript debugging requires version 2.0.0+ of the MarkLogic extension and [MarkLogic 10.0-4+](https://developer.marklogic.com/products/marklogic-server/10.0)._
 
@@ -42,16 +43,28 @@ Note that marklogic.documentsDb *must* be declared in order to attach to remote 
 You can also set `marklogic.authType` to `DIGEST` or `BASIC`. Digest is the default,
 and works even if the server is running basic authentication.
 
-### Connect and query
 
-To evaluate JavaScript
+### Evaluate Queries
 
-1. Type a valid JavaScript query in the editor.
-2. Open the command palette (<kbd>Shift</kbd>+<kbd>Cmd</kbd>+<kbd>P</kbd>)
-3. Select `MarkLogic: Eval JS`
+To evaluate JavaScript, XQuery, SQL, or SPARQL:
 
-Query results will open in a new document in the current workspace.
-`Eval XQuery`, `Eval SQL` and `Eval SPARQL` work the same way.
+1. Type a valid query in the editor.
+2. Open the command palette (<kbd>Shift</kbd>+<kbd>Cmd</kbd>+<kbd>P</kbd>).
+3. Select `MarkLogic: Eval JS`, `MarkLogic: Eval XQuery`, `MarkLogic: Eval SQL`, or `MarkLogic: Eval SPARQL` - depending on the type of query.
+
+Query results will apper in the `MLXPRS: RESULTS` tab in the bottom panel, or open in a new editor tab - depending on the value of the `Marklogic: Results In Editor Tab` setting.
+
+
+### Submit Optic Queries
+
+To run an Optic query (either DSL or serialized):
+
+1. Type a valid query in the editor.
+2. Open the command palette (<kbd>Shift</kbd>+<kbd>Cmd</kbd>+<kbd>P</kbd>).
+3. Select one of the `MarkLogic: Submit Optic query - <Response Format>` commands, depending on the desired response format.
+
+Query results will apper in the `MLXPRS: RESULTS` tab in the bottom panel, or open in a new editor tab - depending on the value of the `Marklogic: Results In Editor Tab` setting.
+
 
 ### Inspect a module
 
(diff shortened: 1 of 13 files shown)


README.md
```diff
@@ -12,8 +12,9 @@ _Develop, run, and debug code for MarkLogic in the popular VS Code IDE_
 
 * Syntax highlighting and IntelliSense for MarkLogic Server-Side JavaScript and XQuery
 * Interactive debugging of JavaScript and XQuery running in MarkLogic, including attaching to in-flight requests and inspecting live variables
-* Real-time query evaluation of JavaScript, XQuery, SQL, and SPARQL against a live Data Hub Service or MarkLogic instance
+* Real-time query evaluation of JavaScript, XQuery, SQL, SPARQL, and Optic against a MarkLogic instance
 * View modules (read-only) in the editor
+* Run MarkLogic Unit Tests
```
rjrudin: Make this a link to the marklogic-unit-test repository, since I imagine a typical developer wouldn't know what this refers to. 

And per the text conversation, I think we should do "Run marklogic-unit-test module", which also abstracts away whether one or many tests are being run (marklogic-unit-test is a bit loose about what exactly a "test" refers to, but we know exactly what a "module" refers to).

README.md
```diff
@@ -237,6 +262,13 @@ Connecting a server will automatically pause all requests so that you can attach
 
 **Note: Only requests that are launched after a server is connected/made debug server can be attached.**
 
+Here are a couple curl commands that can be used with the test-app to start sample requests:
```
rjrudin: I don't think we should include any examples in the README that depend on the test-app. The test-app is for mlxprs developers and should only be referenced in the CONTRIBUTING. If we need examples that depend on an application being deployed to MarkLogic, we should have an e.g. "./examples/(some name)" project that the README provides instructions on deploying and trying out. (This is also out of scope of the JIRA ticket and thus better addressed via a separate ticket/PR).

client/marklogicUnitTestClient.ts
```diff
@@ -0,0 +1,85 @@
...
+            authType: String(cfg.get('marklogic.authType')).toUpperCase(),
+            ssl: Boolean(cfg.get('marklogic.ssl')),
+            pathToCa: String(cfg.get('marklogic.pathToCa') || ''),
+            rejectUnauthorized: Boolean(cfg.get('marklogic.rejectUnauthorized'))
+        };
+        const newParams = new MlClientParameters(configParams);
+        return new ClientContext(newParams);
+    }
+
+    public async sendTestFile(
```
rjrudin: This seems like it should be named `runTestModule` - the current name implies that a file is being sent, which isn't the case. Rather, we're invoking the marklogic-unit-test REST endpoint for running precisely one test module.

client/marklogicUnitTestClient.ts
```diff
@@ -0,0 +1,85 @@
...
+        cfg: WorkspaceConfiguration, editor: TextEditor
+    ): Promise<void> {
+        const splitString = 'test/suites/';
+        const filePath = editor.document.uri.path;
+        const splitLocation = filePath.lastIndexOf(splitString) + splitString.length;
+        if (splitLocation > splitString.length) {
```
rjrudin, line 65: If this is not the case - which I assume could happen if I run this command on a file that is not under "test/suites" - then nothing happens, correct? I think we'd want a message explaining why nothing happened - e.g. "Unable to run test; please ensure the file is a marklogic-unit-test module located in a test suite directory". (I'm also wondering if that will work on Windows because of the use of forward slashes).

client/marklogicUnitTestClient.ts
```diff
@@ -0,0 +1,85 @@
...
+
+            const dbClientContext: ClientContext = this.newMarklogicUnitTestClient(cfg);
+            requestMarkLogicUnitTest(dbClientContext, testSuite, testFile)
+                .result((testResults: string) => {
+                    const testResultsHtml = MlxprsWebViewProvider.convertXmlResponseToHtml(testResults);
+                    MarkLogicUnitTestClient.mlxprsWebViewProvider.updateViewContent(testResultsHtml);
+                    return null;
+                })
+                .catch(error => {
+                    const mlxprsError: MlxprsError = buildMlxprsErrorFromError(error, `Unable to run the test(s): ${error['code']}`);
```
rjrudin: Change the message to "Unable to run the test module" as part of standardizing on "run test module" or "run marklogic-unit-test module".

README.md
```diff
@@ -63,6 +76,18 @@ To view a module from the configured modules database:
 
 The module will appear read-only in a new text buffer.
 
+
+### Running MarkLogic Unit Tests
+
+This plugin provides a convenient method for running [MarkLogic Unit Tests](https://marklogic-community.github.io/marklogic-unit-test/) within your MarkLogic server. To get started, your test suites and files must be organized under a "test/suites" directory. Additionally, you need to set the `Marklogic: Unit Test Port` setting to the port number of the App Server that can run your unit tests. Finally, to run a test file:
```
rjrudin: Re: where the test modules should be - unfortunately, the marklogic-unit-test docs are not clear about this - https://marklogic-community.github.io/marklogic-unit-test/docs/conventions/ . In fact, the guidance of using "src/test/suites" is most likely not going to work. In practice, the correct path will usually be "src/test/ml-modules/root/test/suites". 

The only good advice from the docs is in the "Getting Started" page - https://marklogic-community.github.io/marklogic-unit-test/ - where it recommends looking at ml-gradle's example project, which does show the correct path being used. Sigh. 

Clearly the marklogic-unit-test docs need to be improved. Until then, I think these docs should go with what's going to work the majority of the time, which is to do the following:

1. State that the files should be under "src/test/ml-modules/root/test/suites".
2. Point the user to the ml-gradle example project for an example of how to setup the project to use marklogic-unit-test. 

Ideally, we'll improve the marklogic-unit-test docs soon to make it crystal clear how to setup your project, and then the mlxprs docs can simply refer to that.

package.json
```diff
@@ -326,20 +331,26 @@
                         "default": 8000,
                         "description": "The port of a REST API app server to connect to"
                     },
-                    "marklogic.documentsDb": {
+                    "marklogic.unitTestPort": {
```
rjrudin: For property name, let's just do "testPort" with a description of e.g. "Optional port for a REST API server that supports running marklogic-unit-test tests". This is based on ml-gradle using the names `mlRestPort` and `mlTestRestPort`, which a developer is probably making use of when using marklogic-unit-test. mlxprs doesn't use the word "rest" in the property names, which is fine, so I think "test port" would be the most consistent name to use between the two tools.

rjrudin (changes requested):

README.md
BillFarber: Removed

client/marklogicUnitTestClient.ts
BillFarber: I got caught up in making it similar to the "eval" functions. That was stupid name.
Fixed.

client/marklogicUnitTestClient.ts
BillFarber: Added in code to get the OS path separator and it produces an error message (as suggested) when the expected path does not exist.

README.md
```diff
@@ -63,6 +76,18 @@ To view a module from the configured modules database:
 
 The module will appear read-only in a new text buffer.
 
+
+### Running MarkLogic Unit Test Modules
```
rjrudin: Do "marklogic-unit-test" as that's the official name (though to keep users on their toes, https://github.com/marklogic-community/marklogic-unit-test uses both "marklogic-unit-test" and "MarkLogic unit test"). 

Also, there are a few other "MarkLogic Unit Test" references in the next paragraph to change to "marklogic-unit-test".

Finally, change "MarkLogic: Unit Test Port" to "MarkLogic: Test Port".

package.json
```diff
@@ -101,6 +102,10 @@
             {
                 "command": "extension.showModule",
                 "title": "MarkLogic: Show module"
+            },
+            {
+                "command": "extension.runTestModule",
+                "title": "MarkLogic: Run MarkLogic Unit Test"
```
rjrudin: Should do "MarkLogic: Run marklogic-unit-test module".

PR #220, opened 2023-07-18
DEVEXP-397: Validate TDEs

BillFarber:
Also updates the package-lock files to take care of a medium vulnerability in semver.


client/extension.ts
```diff
@@ -233,3 +244,20 @@ export function deactivate(context: vscode.ExtensionContext): void {
 function handleUnload(context: vscode.ExtensionContext, arrayOfUnloadableObject: vscode.Disposable[]) {
     arrayOfUnloadableObject.forEach(unloadableObject => context.subscriptions.push(unloadableObject as vscode.Disposable));
 }
+
+function newMarklogicTdeValidateClient(cfg: vscode.WorkspaceConfiguration): ClientContext {
+    const configParams: Record<string, unknown> = {
```
rjrudin, line 249: One refactoring idea - I think this is the 4th place where this code exists - i.e. a Record being created based on WorkspaceConfiguration properties. Could this become a new constructor in MlClientParameters so that the various property names don't need to be repeated?

I'm specifically thinking for when `cloudApiKey` is added - the fewer places that need to be modified to support that, the better.

client/marklogicTdeValidateClient.ts
```diff
@@ -0,0 +1,127 @@
...
+        dbClientContext: ClientContext, tdeText: string
+    ): Promise<unknown> {
+        const tdeFormat = this.getTdeFormat(tdeText);
+
+        if (tdeFormat === 'JSON') {
+            return this.validateJsonTde(dbClientContext, tdeText);
+        } else if (tdeFormat === 'XML') {
+            return this.validateXmlTde(dbClientContext, tdeText);
+        } else {
+            const errorMessage = 'To be validated, the template must be either valid JSON or XML';
```
rjrudin: Toss a period at the end since it's a sentence.

client/marklogicTdeValidateClient.ts
```diff
@@ -0,0 +1,127 @@
...
+
+    private handleResults(validationResults: object): string {
+        const result = validationResults[0]['value'] as string;
+        const formattedResults = MlxprsWebViewProvider.convertTextResponseToHtml(JSON.stringify(result, null, 2));
+        MarkLogicTdeValidateClient.mlxprsWebViewProvider.updateViewContent(formattedResults);
+        return result;
+    }
+
+    private handleError(error: Error): string {
+        const errorMessage = `Unable to validate the module: ${error['body']['errorResponse']['message']}`;
```
rjrudin: I think "module" (which normally refers to a file of code) should be replaced with "template".

client/marklogicTdeValidateClient.ts
```diff
@@ -0,0 +1,127 @@
...
+
+    public constructor(context: ExtensionContext) {
+        this.extensionContext = context;
+    }
+
+    private getTdeFormat(tdeText: string) {
+        let tdeFormat = null;
+        try {
+            JSON.parse(tdeText);
+            tdeFormat = 'JSON';
```
rjrudin: Looks like you can just return "JSON" here if this succeeds instead of trying to parse the template as XML too.

package.json
```diff
@@ -106,6 +106,10 @@
             {
                 "command": "extension.runTestModule",
                 "title": "MarkLogic: Run marklogic-unit-test module"
+            },
+            {
+                "command": "extension.validateTde",
+                "title": "MarkLogic: Validate TDE"
```
rjrudin: Let's call it "Validate TDE template", as it's a "template" that's being validated.

rjrudin (changes requested):

client/extension.ts
BillFarber: Created a separate task for this: https://project.marklogic.com/jira/browse/DEVEXP-512

PR #221, opened 2023-07-19
DEVEXP-512: Refactor ClientParameter builders

BillFarber:
Also fixed a small bug with the TDE template validation.

### DIFF
--- a/client/marklogicClient.ts
+++ b/client/marklogicClient.ts
@@ -18,6 +18,7 @@
 
 import * as fs from 'fs';
 import * as ml from 'marklogic';
+import { WorkspaceConfiguration } from 'vscode';
 
 import { MlxprsStatus } from './mlxprsStatus';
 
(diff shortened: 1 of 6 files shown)


client/marklogicClient.ts
```diff
@@ -407,3 +408,20 @@ export function newMarklogicManageClient(dbClientContext: ClientContext, manageP
     manageClientParams.modulesDb = null;
     return new ClientContext(manageClientParams);
 }
+
+export function newClientParamsFromWorkspaceConfigurationAndOverrides(cfg: WorkspaceConfiguration, overrides: object = {}): MlClientParameters {
```
rjrudin: I think `newClientParams` is fine here, until there's another instance of the function that would require a name that distinguishes it from that other function. But we can cross that bridge later.

rjrudin (changes requested):
Looks good, just think a shorter name is good for now.

PR #223, opened 2023-07-21
DEVEXP-398: Provide Node Extract from TDE

BillFarber:
Also updates a couple of transitive dependencies.

### DIFF
--- a/client/marklogicTdeValidateClient.ts
+++ b/client/marklogicTdeValidateClient.ts
@@ -26,6 +26,12 @@ import { MlxprsWebViewProvider } from './mlxprsWebViewProvider';
 
 export class MarkLogicTdeValidateClient {
     static mlxprsWebViewProvider: MlxprsWebViewProvider = null;
+    static xmlParserOptions = {
+        ignoreAttributes: false,
+        attributeNamePrefix: '@_',
+        format: true
+    };
+    static xmlParser = new XMLParser(MarkLogicTdeValidateClient.xmlParserOptions);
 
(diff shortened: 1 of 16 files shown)


anu3990 (approved):

client/marklogicTdeValidateClient.ts
```diff
@@ -127,4 +125,96 @@ export class MarkLogicTdeValidateClient {
             return errorMessage;
         }
     }
+
+    private buildDocumentList(uriVarVal: string): string {
+        const valArray = uriVarVal.split(',');
```
rjrudin: This isn't 100% reliable because MarkLogic allows for a URI to have a comma in it. You could come up with another delimiter that isn't allowed in a URI - but I would hold off on this until someone asks for it. I think if someone does want to test multiple existing URIs, the current UX will be fine - i.e. test one, analyze the results; then test another by changing the MLXPRS_TEST_URI value and analyze the results. Or - just write the equivalent of that in an editor and eval it via mlxprs.

package.json
```diff
@@ -110,6 +110,10 @@
             {
                 "command": "extension.validateTdeTemplate",
                 "title": "MarkLogic: Validate TDE template"
+            },
+            {
+                "command": "extension.tdeExtractNodes",
+                "title": "MarkLogic: Extract data from TDE"
```
rjrudin: How about "Extract data via TDE" - "from TDE" to me sounds like I'm extracting data from the TDE itself, but I'm really using the TDE to extract data from some other thing.

PR #227, opened 2023-07-27
DEVEXP-264: Submit a GraphQL query

BillFarber:


anu3990 (approved):

package.json
```diff
@@ -83,6 +83,10 @@
                 "command": "extension.sendRowsXmlQuery",
                 "title": "MarkLogic: Submit Optic query - return XML"
             },
+            {
+                "command": "extension.sendGraphQlQuery",
+                "title": "MarkLogic: Submit GraphQL Query"
```
rjrudin, line 88: Can address this in a follow-up PR - there's some existing inconsistency with capitalization - i.e. "Optic query" above, and "GraphQL Query" here. 

Based on OOTB VSCode commands, it looks like Camel Case is the way to go. So this is good, but should make a PR to adjust the other commands that don't use it (though I would leave "marklogic-unit-test" as "marklogic-unit-test" since that's its name). 

I think these are the ones to capitalize:

1. All the "Submit Optic Query" ones.
2. "Show Module"
3. "Run marklogic-unit-test Module"
4. "Validate TDE Template"
5. "Extract Data Via TDE

PR #231, opened 2023-08-03
DEVEXP-529: running an ML Unit Test fails when a content DB is defined

BillFarber:


anu3990 (approved):

client/marklogicClient.ts
```diff
@@ -393,13 +393,21 @@ export async function filterJsServerByConnectedStatus(
...
+    const endpoint = '/v1/resources/marklogic-unit-test';
+    let endpointParameters = `rs:format=xml&rs:func=run&rs:suite=${encodeURIComponent(testSuite)}`;
     if (testFile) {
-        endpoint = `${endpoint}&rs:tests=${testFile}`;
+        endpointParameters = `${endpointParameters}&rs:tests=${encodeURIComponent(testFile)}`;
     }
     return dbClientContext.databaseClient.internal.sendRequest(
         endpoint,
         (requestOptions: ml.RequestOptions) => {
+            // The marklogic-unit-test parameters must be added here since the Node Client automatically
```
rjrudin: I think it's fine to remove this comment - the reason why there's a "?" doesn't really matter to mlxprs, as what's going on below the comment is common and makes sense.

rjrudin (approved):

PR #232, opened 2023-08-03
DEVEXP-531: Clarify README with regard to the "debug-any-requests" privilege

BillFarber:
Also, standardized "app server" to "App Server" in the file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -309,10 +309,10 @@ Here's an example of _attach_ configurations for JavaScript and XQuery:
     }
 ```
 
-Attach mode intercepts a paused request in a given *debug server*, an app server connected to the VS Code debugger. To connect to an app server for debugging:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -374,11 +374,13 @@ To run queries with the MarkLogic JavaScript and XQuery Debugger, a user will ne
 - **xdmp-eval-modules-change**: to use a non-default modules database or modules root
 - **xdmp-eval-modules-change-file**: to use the filesystem for modules
 
-For debugging, a user must also have at least one of these privileges:
+For debugging, a user must also have at least one of these privileges to evaluate a JavaScript or XQuery module in debug mode.
 
 - **debug-my-request**: for debugging requests launched by the debug user only
```
rjrudin: Should be "requests" instead of "request", which also applies to the sentence below it.

rjrudin (changes requested):

PR #237, opened 2023-08-11
Fixes to the XQY Debug Adapter to permit automated testing

BillFarber:
The primary problem was console.debug statements causing problems with the communications protocol between VSCode and the Debug Adapter. I'm assuming they have served a purpose, so I'm simply commenting them out for now.

Added a new XQY Debug test to verify the capability.

Also, refactored some of the JavaScript Debug Adapter to make the differences clear.

Many minor & patch version updates.


rjrudin (approved):
Looks good, just thinking the commented out stuff should either be removed or another comment should be added to explain why all these commented out console.debug statements are being left in.

PR #241, opened 2023-08-21
New user docs - outline

BillFarber:
This is just a rough start to get the ball rolling for the new style of user docs.


client/extension.ts
```diff
@@ -24,6 +24,14 @@ import { LanguageClient, LanguageClientOptions, ServerOptions, TransportKind } f
 import { ClientResponseProvider } from './clientResponseProvider';
 import { ConfigurationManager } from './configurationManager';
 import { EditorQueryType, EditorQueryEvaluator } from './editorQueryEvaluator';
+import { ClientContext } from './marklogicClient';
```
rjrudin: Is this a fix for something, or was it not meant to be included? Same goes for the xqyFormatting changes too. Was only expecting files in the ./docs directory.

rjrudin (commented):
This is very helpful to see and a good start. Looking at the different pages, I'm thinking a "parent" page could be used - like the "Managing Documents" page at https://marklogic.github.io/marklogic-python-client/documents or "Getting Started" at https://marklogic.github.io/marklogic-spark-connector/docs/getting-started - to group some of these features. And the groups would be different kinds of "support" that mlxprs offers.

Spitballing (not necessarily in this order - though perhaps a good order would be Development, Query, Testing, Debugging):

- "Debugging Support" - with child pages for JavaScript and XQuery, as my impression is that each is different and would benefit from a dedicated page? 
- "Testing Support" - a single page that describes integration with marklogic-unit-test.
- "Query Support" - child pages for Optic, GraphQL, SQL, and SPARQL, with a theme of "mlxprs helps you develop and verify features involving the different query languages supported by MarkLogic by allowing you to easily execute ad hoc queries and see the results."
- "Development Support" - I was initially thinking TDE would go under "Query Support", but perhaps it makes more sense here because the goal is to develop a TDE. And this group could also have child pages for formatters, syntax highlighters, intellisense, etc - however it makes sense to group those together.

client/extension.ts
BillFarber: ugh, no.
My rebase didn't work as expected.

BillFarber: Those changes have been removed.

docs/index.md
mitchshepherd, line 1: Remove the first comma on line 7 after "Code"

BillFarber:
Update Outline:
image

docs/index.md
```diff
@@ -0,0 +1,17 @@
...
+The MLXPRS: MarkLogic extension for Visual Studio Code helps you to develop, run, and debug code for MarkLogic. The following features improve developer efficiency by giving the developer the ability to perform frequently used functions in one place, without leaving the IDE.
+
+
+## Features
+
+* Syntax highlighting and IntelliSense for MarkLogic Server-Side JavaScript and XQuery
+* Interactive debugging of JavaScript and XQuery running in MarkLogic, including attaching to in-flight requests and inspecting live variables
+* Real-time query evaluation of JavaScript, XQuery, SQL, SPARQL, Optic, and GraphQL against a MarkLogic instance
+* View modules (read-only) in the editor
+* Run [marklogic-unit-test module](https://github.com/marklogic-community/marklogic-unit-test)
```
rjrudin: module -> modules . Also, while a link is helpful here, I think we either want no links or we want each of these to link to something useful - like TDE could link to the TDE docs, SPARQL/Optic to their respective docs, etc. I am thinking "no links" is the way to go here - each page specific to a topic can have appropriate links in it.

docs/debbuging-support/debugging-support.md
rjrudin, line 1: Typo - "debugging-support" -> "debugging-support".

docs/index.md
```diff
@@ -0,0 +1,17 @@
+---
+layout: default
+title: Introduction
+nav_order: 1
+---
+
+The MLXPRS: MarkLogic extension for Visual Studio Code helps you to develop, run, and debug code for MarkLogic. The following features improve developer efficiency by giving the developer the ability to perform frequently used functions in one place, without leaving the IDE.
```
rjrudin: I think "develop, run, and debug code" was a reasonable description in the early days of mlxprs, but that scope seems too small now - mlxprs can do a lot more and isn't just for developer writing custom code. How about this:

"MLXPRS is an extension for Visual Studio Code that simplifies and speeds up the development of applications for MarkLogic.  Developers can enjoy all of the benefits that they're accustomed to with Visual Studio Code while leveraging the following features for their MarkLogic applications:"

And then remove the "Features" heading and just put the bulleted list after that sentence that ends with the colon.

docs/installationAndConfiguration.md
```diff
@@ -0,0 +1,28 @@
+---
+layout: default
+title: Installation & Configuration
+nav_order: 2
+---
+
+Install this tool using the VS Code built-in [marketplace](https://marketplace.visualstudio.com/items?itemName=mlxprs.mlxprs). Search “MarkLogic” from the Extension tab of the activity bar. Click “Install” to download and install the extension.
```
rjrudin: Replace "tool" with "extension" - we should exclusively use "extension" when referring to mlxprs.

docs/installationAndConfiguration.md
```diff
@@ -0,0 +1,28 @@
...
+
+
+
+The MarkLogic extension may be configured in the VS Code Settings. You can also change the configuration settings just for the workspace by adding the values to the `settings.json` file in the .vscode directory of your workspace.
```
rjrudin: I think this is worth explaining - I often forget how to actually get to the settings. The approach I just used was choosing "Preferences: Open User Settings" and then we should tell the user to look under "Extensions" and then "MLXPRS".

docs/installationAndConfiguration.md
```diff
@@ -0,0 +1,28 @@
...
+
+```json
```
rjrudin, line 13: This comment is out-of-scope of this PR, but we'll eventually want a table that lists all the possible settings and what appropriate values are.

docs/query-and-script-support/ExecutingSqlAndSparqlQueries.md
rjrudin, line 1: I think we want separate pages for Optic, GraphQL, SQL, and SPARQL. They're all a bit different and worthy of their own page, even if some of the pages are quite short. SQL and SPARQL don't really have anything in common either.

docs/query-and-script-support/evaluatingModules.md
rjrudin, line 1: What's the difference between "evaluating js/xquery modules" and "evaluating queries and scripts"? Aren't both of those describing "Eval JS" and "Eval XQuery"?

BillFarber: No *real* difference. But they feel different to me.

rjrudin (approved):

PR #245, opened 2023-12-29
Update the README with newly discovered debugging limitations.

BillFarber:


README.md
```diff
@@ -361,6 +361,12 @@ Evaluating variables in module scope may throw reference error. Possible workaro
 - use the variables of interest inside a function, and inspect
 - place an `eval()` statement inside the affected module
 
+While debugging an MJS file, variables and watch expressions do work work as expected.
+
+Resource Services do not pause as expected and therefore cannot currently be debugged.
```
rjrudin: Wrap "Resource Services" with a link to https://docs.marklogic.com/guide/rest-dev/extensions#id_59188 so it's clear what that refers to.

README.md
```diff
@@ -361,6 +361,12 @@ Evaluating variables in module scope may throw reference error. Possible workaro
...
+
+There have been unverified reports of breakpoints not working properly when running VS Code in a Windows environment.
```
rjrudin: I don't think it helps a user to be told this, as it reads as "Hey, this may or may not work in Windows" - instead, I think we should wrap this section up with "If you encounter any issue with the debugger that is not addressed above, please [open a ticket](https://github.com/marklogic/mlxprs/issues) with steps for reproducing the issue."

rjrudin (changes requested):

rjrudin (approved):

PR #253, opened 2024-02-16
MLE-4107 Can now view MarkLogic information in a sidebar activated by the activity bar

BillFarber:


BillFarber:
I realized that I have done anything about potential errors, so I've converted this to a draft PR for the time-being.

client/extension.ts
```diff
@@ -119,28 +121,36 @@ export function activate(context: vscode.ExtensionContext): void {
         const cfg: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration();
         const dbClientContext: ClientContext = getDbClientWithoutOverrides(cfg, context.globalState);
         if (dbClientContext) {
-            JsDebugManager.connectToJsDebugServer(dbClientContext);
+            JsDebugManager.connectToJsDebugServer(dbClientContext).then(() => {
+                vscode.commands.executeCommand('markLogicServerStatus.refreshEntry');
```
rjrudin: This line looks like a function waiting to be extracted, as it appears to be the same in several places - e.g. `refreshServerStatus` (though I'm wondering if it should reference the actual GUI component, which I think would be more self-documenting - e.g. `refreshServerStatusInActivityBar`?).

client/extension.ts
```diff
@@ -230,6 +240,11 @@ export function activate(context: vscode.ExtensionContext): void {
     JsDebugManager.registerMlxprsStatusBarItem(mlxprsStatus);
     mlxprsStatus.requestUpdate();
 
+    // Add a MarkLogic ActivityBar entry
```
rjrudin: I'm wondering about the naming here. The server status is shown in a VSCode "tree data provider", so the variable name below makes sense. Does calling `registerTreeDataProvider` automatically add it to what VSCode calls the "activity bar"?

client/marklogicClient.ts
```diff
@@ -221,6 +245,63 @@ return $json-servers
...
+                .result((result: ServerListResponse) => {
+                    resolve(result);
+                })
+                .catch((error) => {
+                    reject(error);
+                });
+        });
+    }
+
+    async executeGenericGetRequest(url: string): Promise<string> {
```
rjrudin: Did you mean to reuse this above? I saw the duplication in getDatabaseList and getAppServerList and was thinking - looks like there's a generic GET function to be extracted that just accepts a path - and that seems to be this function.

client/marklogicServerStatus.ts
```diff
@@ -0,0 +1,205 @@
...
+        ];
+    }
+
+    private async buildDatabaseEntries(): Promise<MarkLogicServerStatus[]> {
+        return this.dbManageClientContext.getDatabaseList()
+            .then((databaseResponse) => {
+                const serverStatusList: MarkLogicServerStatus[] = [];
+                const responseList = databaseResponse['database-default-list']['list-items']['list-item'];
+                responseList.forEach(databaseItem => {
+                    const url = `http://localhost:8001/database-admin.xqy?section=database&database=${databaseItem.idref}`;
```
rjrudin: This URL is hardcoded, and presumably can have the scheme, host, and post all replaced (I forget if mlxprs supports SSL, but assuming it does, you'd need to determine if you want "http" or "https" here).

client/extension.ts
BillFarber: Are you talking about all 3 of these lines repeating?
```
            JsDebugManager.connectToJsDebugServer(dbClientContext).then(() => {
                vscode.commands.executeCommand('markLogicServerStatus.refreshEntry');
            });
```
If so, they don't quite repeat. That first line is a bit different each time.

rjrudin (approved):
Looks good!

PR #256, opened 2024-02-20
Initial basePath work - probably the first of several

BillFarber:
Gets some plumbing in place,
Adds Rob's code for a reverse proxy for testing
Gets the Server Status view to work with the proxy.


client/marklogicClient.ts
```diff
@@ -73,6 +73,12 @@ export class MlClientParameters {
     ssl: boolean;
     pathToCa: string;
     rejectUnauthorized: boolean;
+    useProxy: boolean;
+    basePath: string;
+    managePath: string;
```
rjrudin: Include "BasePath" as the suffix for these - that's consistent with ml-gradle and avoids any confusion over what kind of "path" it is.

client/marklogicClient.ts
```diff
@@ -474,9 +489,12 @@ export function requestMarkLogicUnitTest(
...
 }
 
-export function newMarklogicManageClient(dbClientContext: ClientContext, managePort: number): ClientContext {
+export function newMarklogicClientWithPort(
+    dbClientContext: ClientContext, targetPort: number, targetPath: string
+): ClientContext {
     const manageClientParams: MlClientParameters = JSON.parse(JSON.stringify(dbClientContext.params));
-    manageClientParams.port = managePort;
+    manageClientParams.port = targetPort;
+    manageClientParams.restPath = targetPath;
```
rjrudin: IIUC, I think this should be `basePath` and not `restPath`?

client/marklogicClient.ts
```diff
@@ -474,9 +489,12 @@ export function requestMarkLogicUnitTest(
     );
 }
 
-export function newMarklogicManageClient(dbClientContext: ClientContext, managePort: number): ClientContext {
+export function newMarklogicClientWithPort(
```
rjrudin, line 488: Something to consider if you may need to add a 3rd argument here in the future - allow the caller to pass in an `overrides` object that lets the caller override whatever is in the `dbClientContext`. That means when a 3rd arg is needed, you don't have to change the method signature.

Example of a caller:

    const client = newMarkLogicClient(this.dbClientContext, {"port": somePort, "basePath": somePath});

And then the function would use those overrides to modify the clientParams.. 

The only downside would be that the callers know the names of "port" and "basePath", but that seems reasonable since those are defined in an interface.

rjrudin (commented):
Looks good, just a couple small things (assuming that `useProxy` is dropped and that the Java Client's reverse proxy is reused).

client/marklogicClient.ts
BillFarber: There's actually already a function to support that right below this one. Perhaps I'll change it.

client/test/integration/markLogicIntegrationTestHelper.ts
```diff
@@ -62,7 +64,9 @@ export class IntegrationTestHelper {
     private clientDefaults = {
         host: this.hostname,
         port: this.port,
+        restPath: this.restPath,
```
rjrudin: Wondering if this is supposed to be "restBasePath" - I didn't find any references to "restPath" in the rest of the PR.

rjrudin (approved):
Looks good, just one had one question.

client/test/integration/markLogicIntegrationTestHelper.ts
BillFarber: fixed

PR #258, opened 2024-02-21
Debug Status view now uses basePath

BillFarber:
Additionally, this has the ripple effect of requiring the mlxprs status bar component to use basePath as well.
Also, required updated some integration tests.

### DIFF
--- a/client/marklogicClient.ts
+++ b/client/marklogicClient.ts
@@ -174,8 +174,14 @@ export class ClientContext {
     }
 
 
-    async getConnectedServers(requestingObject: MlxprsStatus | null, managePort: number, updateCallback: (connectedServers: string[]) => void): Promise<string[]> {
-        const connectedJsServers: MlxprsQuickPickItem[] = await getFilteredListOfJsAppServers(this, managePort, 'true');
+    async getConnectedServers(
+        requestingObject: MlxprsStatus | null,
+        managePort: number,
+        manageBasePath: string,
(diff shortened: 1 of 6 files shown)


client/marklogicClient.ts
```diff
@@ -486,14 +497,16 @@ export function requestMarkLogicUnitTest(
...
-    const manageClientParams: MlClientParameters = JSON.parse(JSON.stringify(dbClientContext.params));
-    manageClientParams.port = targetPort;
-    manageClientParams.restBasePath = targetPath;
-    manageClientParams.contentDb = null;
-    manageClientParams.modulesDb = null;
-    return new ClientContext(manageClientParams);
+    const newClientParams: MlClientParameters = JSON.parse(JSON.stringify(dbClientContext.params));
+    newClientParams.port = targetPort;
+    newClientParams.restBasePath = targetPath;
+    newClientParams.contentDb = null;
```
rjrudin, line 511: I know this code was already here and this is just a rename, but do you know why `contentDb` and `modulesDb` are nulled out? I think that deserves a comment as it seems surprising - i.e. to a casual reader, I'm left wondering "Wait, the user may have configured these, so why are we now ignoring them?"

rjrudin (approved):
Looks good, just wondering if a comment would be a good addition in one existing place (i.e. a comment may have been helpful prior to this PR, the PR didn't introduce the need for a comment).

PR #259, opened 2024-02-21
Can now debug behind a proxy server

BillFarber:
Also working:
Connect/Disconnect to JS & XQY servers.
Eval JS, XQY, GraphQl, Optic, Optic DSL, SPARQL, and SQL.
Validate TDE template.
Extract data via TDE.
Show module from database.

### DIFF
--- a/client/JSDebugger/mlRuntime.ts
+++ b/client/JSDebugger/mlRuntime.ts
@@ -118,6 +118,7 @@ export class MLRuntime extends EventEmitter {
         this._password = args.password;
         this._dbgPort = args.managePort;
         this.managePort = args.managePort;
+        this.manageBasePath = args.manageBasePath;
         this._ssl = args.ssl;
         this._scheme = this._ssl ? 'https' : 'http';
         this._rejectUnauthorized = args.rejectUnauthorized;
@@ -129,6 +130,7 @@ export class MLRuntime extends EventEmitter {
             new MlClientParameters({
                 host: this._hostName,
                 port: this._dbgPort,
+                restBasePath: this.manageBasePath,
                 user: this._username,
                 pwd: this._password,
                 contentDb: args.database,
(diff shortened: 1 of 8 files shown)


client/JSDebugger/mlRuntime.ts
```diff
@@ -129,6 +130,7 @@ export class MLRuntime extends EventEmitter {
             new MlClientParameters({
                 host: this._hostName,
                 port: this._dbgPort,
+                restBasePath: this.manageBasePath,
```
rjrudin, line 133: Is this correct? I see how `MlClientParameters` has both `restBasePath` and `manageBasePath` properties - this seems odd because I would expect these two values to be different.

BillFarber: It's definitely convoluted, but it is correct.

MlClientParameters is capable of holding all of the MLXPRS settings individually.
However, when it is used to build a ClientContext object, only the "port" and "restBasePath" are used. The other ports/paths are ignored. So, I need to set "port" and "restBasePath" to the values I need for this specific client.
In this case, I need a client using the "manage" parameters.

One refactor that I've been considering - In the constructor for ClientContext, add a second parameter specifying which port/path pair to use. That would at least clean up this wierdness.

rjrudin: Got it - I think the confusion for me is I assume that `MlClientParameters` contains the params for one client object. The "mlxprs config" would have multiple ports and paths - one pair for rest/test/manage/admin - but when a `ClientParams` object is created, it should have `port` and `basePath` properties, and the caller chooses which pair of values to pass in. 

I don't know how feasible it would be to refactor `MlClientParameters` to be like that, but it would avoid confusion like `restBasePath=manageBasePath`.

rjrudin (approved):

PR #261, opened 2024-02-22
First step towards replacing MlClientParameters

BillFarber:
MlClientParameters does too much now tracking different ports and paths, resulting in convoluted code. This is a first step in fixing that class. Replacing MlClientParameters in one fell swoop is too big and would lead to a ridiculous PR.

MlConnectionConfigurator is intended to be one stop shopping for storing all ML connection configuration parameters. Then using those values to generate any type of ClientContext required.

Note - this is part PR, part design proposal.
If you like it, we'll go with it.
If you hate it, I'll scrap it.
If you think it's interesting but need's more design, we'll discuss it.

### DIFF
--- a/client/XQDebugger/xqyDebugConfigProvider.ts
+++ b/client/XQDebugger/xqyDebugConfigProvider.ts
@@ -23,12 +23,13 @@ import {
 
 import { MlxprsErrorReporter } from '../mlxprsErrorReporter';
 import { MlClientParameters } from '../marklogicClient';
+import { buildClientFactoryFromConfigurationManager } from '../clientFactory';
 import { MlxprsError } from '../mlxprsErrorBuilder';
 import { XqyDebugManager, DebugStatusQueryResponse } from './xqyDebugManager';
-import { ConfigurationManager } from '../configurationManager';
 
-
+/* eslint-disable @typescript-eslint/no-unused-vars */
 export class XqyDebugConfiguration implements DebugConfiguration {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
     [key: string]: any
     type: string;
     name: string;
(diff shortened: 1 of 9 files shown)


client/mlConnectionConfigurator.ts
```diff
@@ -0,0 +1,207 @@
+import * as ml from 'marklogic';
+
+import { ClientContext, MlClientParameters } from './marklogicClient';
+
+// This class should NOT use the WorkspaceConfiguration (WC) class imported below.
+// The debuggers don't have access to the vscode namespace, including WC.
+// So if WC is used in the class, the debuggers won't be able to use this class.
+// Note that this also precludes the use of ConfigurationManager in the class.
+export class MlConnectionConfigurator implements ml.ConnectionParams {
```
rjrudin: For naming - what about reusing the `ClientFactory` concept from the Java Client? I would try to reuse "client" as much as possible so as not to have confusion over what the different between a "client" and a "connection" is.

client/mlConnectionConfigurator.ts
```diff
@@ -0,0 +1,207 @@
...
+            contentDb: this.database,
+            modulesDb: this.modulesDb,
+            authType: this.authType,
+            ssl: this.ssl,
+            pathToCa: this.pathToCa,
+            rejectUnauthorized: this.rejectUnauthorized,
+        });
+    }
+
+    newMarklogicRestClient(
```
rjrudin: This is good and will help readability a lot. I think this suggests the concepts are:

1. There's a "Configuration" of all the settings that a user can provide.
2. There's a "ClientFactory" that you instantiate, giving it a "Configuration".
3. You then call the appropriate "Make new a Client" Method on the "ClientFactory".

rjrudin (approved):

client/mlConnectionConfigurator.ts
BillFarber: Exactly what I'm going for.

client/mlConnectionConfigurator.ts
BillFarber: I like it.

PR #263, opened 2024-02-23
MLE-4124 Can now use MLXPRS with ML Cloud

BillFarber:
Primarily manual tests


package.json
```diff
@@ -444,20 +444,32 @@
...
                     },
-                    "marklogic.pathToCa": {
+                    "marklogic.apiKey": {
                         "order": 4,
                         "type": "string",
                         "default": "",
+                        "description": "password for MarkLogic connection"
+                    },
+                    "marklogic.accessTokenDuration": {
+                        "order": 5,
```
rjrudin: While I think it's good to make this configurable so that we can tweak the number - which makes testing token renewal far easier - I don't think we should expose it to users as I can't think of a single use case for a user to lower this. Exposing it only adds confusion in my mind because it causes a user to wonder, "Huh, why would I want to change this?"

Also the description is copy/pasted, but that can easily be fixed by removing this from the UI . :)

rjrudin (changes requested):

rjrudin (approved):

PR #267, opened 2024-03-11
Finishing the major refactor

BillFarber:
ClientContext objects are now only created in the ClientFactory.
MlParameters objects are now used appropriately.


client/clientFactoryVsCodeUtil.ts
rjrudin, line 1: I'm wondering about the name here - "VsCodeUtil" doesn't imply the purpose of this class. "clientFactory" makes sense in the above module... maybe just "vscodeClientFactory" here, with a comment stating e.g. "Only use this when you're in a context that has access to the 'vscode' classes."

client/test/suite/client.test.ts
```diff
@@ -112,21 +111,21 @@ suite('Extension Test Suite', () => {
...
+        const firstClient: ClientContext = getDbClient(queryWithDocumentsDbOverride, SJS, config);
         firstClient.params.host = newHost;
 
         // Verify next client is different since firstClient's params were modified
-        const secondClient: ClientContext = getDbClient(queryWithDocumentsDbOverride, SJS, config, gstate);
+        const secondClient: ClientContext = getDbClient(queryWithDocumentsDbOverride, SJS, config);
         assert.notStrictEqual(firstClient, secondClient);
         assert.notDeepStrictEqual(firstClient, secondClient);
 
+        // SKIPPING - we are not currently using a cache
```
rjrudin: Can this be deleted instead of commented out? Seems irrelevant until we decide to use a cache, which may never happen.

client/vscQueryParameterTools.ts
```diff
@@ -73,34 +73,15 @@ export function parseQueryForOverrides(queryText: string, language: string): Rec
  *
  * @returns a MarklogicClient based on the contents of `cfg`
  */
-export function getDbClient(queryText: string, language: string, cfg: WorkspaceConfiguration, state: Memento): ClientContext {
+export function getDbClient(queryText: string, language: string, cfg: WorkspaceConfiguration): ClientContext {
+    /** Currently not using a cache **/
```
rjrudin: This comment seems unnecessary since there's no expectation that we are using a cache anywhere, right?

rjrudin (approved):
Looks good, just a few small comments.

PR #269, opened 2024-03-12
Fixes a bug with refreshing the server status after a bad configuration

BillFarber:


rjrudin (approved):
Looks good, though that function is a good refactor candidate in the future due to have 3 levels of conditionals.

BillFarber:
> Looks good, though that function is a good refactor candidate in the future due to have 3 levels of conditionals.

I considered that, but with handling multiple promises, I like being able to view all the logic in one place.

PR #270, opened 2024-03-12
Can now read docs about the Server Status view

BillFarber:


docs/serverStatusView.md
```diff
@@ -0,0 +1,38 @@
+---
+layout: default
+title: Server Status View
+nav_order: 3
+---
+
+
+### Server Status View
+
+There is a "MarkLogic Server Explorer" icon (
```
rjrudin: I try to avoid passive sentences like "There is" - try an active voice like "mlxprs includes an icon (or button? Not sure of the right term) in the VSCode activity bar, which VSCode displays by default on the left side of the VSCode window. When selected...".

rjrudin (changes requested):

rjrudin (approved):

PR #286, opened 2024-10-21
Fixes a high vulnerability.

BillFarber:
Also updates the docker base image in the test app.
All the automated tests are passing locally, and the manual smoke tests work as well.

I'm wondering if we should have a 3.9.1 release for all of these dependency updates.


rjrudin (approved):
Adjust the commit comment to include a link to the vulnerability, can also put it in the description.

